PR #1, opened 2018-05-14
Base Android project created

lakshyagupta21:


lognaturel:
@lakshyagupta21 @shobhitagarwal1612 I think it's a great idea to work through pull requests and have great commit messages so that the history is easy to follow in the future. However, I'd say that it's fine for @lakshyagupta21 to merge his own pull requests initially to keep things efficient. 

@lakshyagupta21 you can always request code review if there's anything you want specific feedback on and once there's a beta out with users it would be good to do consistently.

What do you think?

shobhitagarwal1612:
Sounds good to me

PR #6, opened 2018-05-18
Hotspot implementation

### LINKED ISSUE #4: WifiHotspot Enable/Disable

### LINKED ISSUE #5: WifiHotspot create/restore configuration

lakshyagupta21:
Close #4 
Close #5 

Tested on Android 6.0, 7.1

This PR includes implementation of different methods of Wifi Hotspot. I've added a button so that hotspot can be enabled and also added notification through which user would be able to know that hotspot is in running state or not.

ezgif com-video-to-gif


shobhitagarwal1612:
@lakshyagupta21 The hotspot generation is working perfectly but there should be a way to stop this as well

shobhitagarwal1612:
Since this PR is related to hotspot implementation, it should have the basic functionalities related to the hotspot:
Enable
Disable
Customise the wifi hotspot configuration
Check if wifi hotspot is enabled/supported,
etc.

shobhitagarwal1612:
I can verify that the enable/disable functions are working fine

share_app/src/main/java/org/odk/share/services/HotspotService.java
```diff
@@ -0,0 +1,160 @@
...
+        state = new HotspotState(this);
+        registerReceiver(stopReceiver, new IntentFilter(ACTION_STOP));
+        startForeground(HOTSPOT_NOTIFICATION_ID, buildForegroundNotification(getString(R.string.hotspot_start), false));
+    }
+
+    @Override
+    public int onStartCommand(Intent intent, int flags, int startId) {
+        String action = intent.getAction();
+
+        if (action.equals(ACTION_START)) {
```
shobhitagarwal1612: this nested `if-else` can be converted to `switch case`

share_app/src/main/java/org/odk/share/controller/WifiHotspot.java
```diff
@@ -0,0 +1,154 @@
...
+    WifiConfiguration lastConfig;
+    WifiConfiguration currConfig;
+    private static final String ssid = "ODK-Share";
+    private static final String TAG = WifiHotspot.class.getClass().getName();
+
+    public WifiHotspot(Context context) {
+        this.context = context;
+        wifiManager = (WifiManager) context.getApplicationContext().getSystemService(context.WIFI_SERVICE);
+        Method[] methods = wifiManager.getClass().getMethods();
+        for (int i = 0; i < methods.length; i++) {
```
shobhitagarwal1612: `for` loop can be replaced with `for-each`

share_app/src/main/java/org/odk/share/controller/WifiHotspot.java
```diff
@@ -0,0 +1,154 @@
...
+
+    Context context;
+    Method getWifiApConfig;
+    Method setWifiApEnable;
+    Method setWifiApConfig;
+    Method isWifiApEnabled;
+    Method getWifiApState;
+    WifiManager wifiManager;
+    WifiConfiguration lastConfig;
+    WifiConfiguration currConfig;
```
shobhitagarwal1612: All these variables should be declared as `private`

share_app/src/main/java/org/odk/share/controller/WifiHotspot.java
```diff
@@ -0,0 +1,154 @@
...
+    private static final String ssid = "ODK-Share";
+    private static final String TAG = WifiHotspot.class.getClass().getName();
+
+    public WifiHotspot(Context context) {
+        this.context = context;
+        wifiManager = (WifiManager) context.getApplicationContext().getSystemService(context.WIFI_SERVICE);
```
shobhitagarwal1612: Use static reference in `context.WIFI_SERVICE` ie `Context.WIFI_SERVICE`

share_app/src/main/AndroidManifest.xml
```diff
@@ -2,20 +2,28 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="org.odk.share">
 
+    <!--<uses-permission android:name="android.permission.CHANGE_NETWORK_STATE" />-->
+    <uses-permission android:name="android.permission.CHANGE_WIFI_STATE" />
+    <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
+    <uses-permission android:name="android.permission.WRITE_SETTINGS" />
```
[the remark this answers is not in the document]
lakshyagupta21: I forget to comment it, it will be needed somewhere in wifi code.

share_app/src/main/java/org/odk/share/controller/WifiHotspot.java
```diff
@@ -0,0 +1,151 @@
...
+
+    private Method getWifiApConfig;
+    private Method setWifiApEnable;
+    private Method setWifiApConfig;
+    private Method isWifiApEnabled;
+    private Method getWifiApState;
+    private WifiManager wifiManager;
+    private WifiConfiguration lastConfig;
+    private WifiConfiguration currConfig;
+    private static final String ssid = "ODK-Share";
```
lakshyagupta21: For now, I've created new wifi configuration with `ssid name= "ODK-Share"` and the hotspot is open(This will be the default config). But when I add the settings there I'll add the option so that user can name the new config and choose the password for the created hotspot.

share_app/src/main/AndroidManifest.xml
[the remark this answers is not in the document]
shobhitagarwal1612: I think only system apps have permission to change settings

share_app/src/main/java/org/odk/share/controller/WifiHotspot.java
shobhitagarwal1612: This is looking great.
Next step would be to pick these configurations from shared preferences and creating a simple settings activity to change these settings.

PR #16, opened 2018-05-20
Screen to show Instances from odk directory

### LINKED ISSUE #10: Create a screen to display instances stored in device under odk directory

lakshyagupta21:
Closes #10 

ezgif com-video-to-gif 1

### DIFF
--- a/share_app/src/main/java/org/odk/share/activities/InstancesList.java
+++ b/share_app/src/main/java/org/odk/share/activities/InstancesList.java
@@ -1,0 +1,67 @@
+package org.odk.share.activities;
+
+
+import android.database.Cursor;
+import android.support.v7.app.AppCompatActivity;
+import android.os.Bundle;
+import android.support.v7.widget.LinearLayoutManager;
+import android.support.v7.widget.RecyclerView;
+
+import org.odk.share.R;
+import org.odk.share.adapters.InstanceAdapter;
+import org.odk.share.dao.InstancesDao;
+import org.odk.share.provider.InstanceProviderAPI;
+
+
+import android.support.v4.app.LoaderManager;
+import android.support.v4.content.Loader;
+import android.support.v7.widget.Toolbar;
+
+import butterknife.BindView;
+import butterknife.ButterKnife;
+
+public class InstancesList extends AppCompatActivity implements LoaderManager.LoaderCallbacks<Cursor> {
+
+    @BindView(R.id.recyclerview) RecyclerView recyclerView;
+    @BindView(R.id.toolbar) Toolbar toolbar;
+
+    protected static final String SORT_BY_NAME_ASC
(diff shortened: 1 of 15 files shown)


share_app/src/main/java/org/odk/share/activities/InstancesList.java
```diff
@@ -0,0 +1,136 @@
...
+
+        LinearLayoutManager llm = new LinearLayoutManager(this);
+        llm.setOrientation(LinearLayoutManager.VERTICAL);
+        recyclerView.setLayoutManager(llm);
+        getSupportLoaderManager().initLoader(INSTANCE_LOADER, null, this);
+    }
+
+
+    @Override
+    public Loader<Cursor> onCreateLoader(int id, Bundle args) {
```
shobhitagarwal1612: This method can be annotated with @NonNull

share_app/src/main/java/org/odk/share/activities/InstancesList.java
```diff
@@ -0,0 +1,136 @@
...
+        return new InstancesDao().getSavedInstancesCursorLoader(SORT_BY_NAME_ASC);
+    }
+
+    @Override
+    public void onLoadFinished(Loader<Cursor> loader, Cursor cursor) {
```
shobhitagarwal1612: `loader` parameter can be annotated with @NonNull

share_app/src/main/java/org/odk/share/activities/InstancesList.java
```diff
@@ -0,0 +1,136 @@
...
+           toggleButton.setText(getString(R.string.select_all));
+           toggleButton.setEnabled(true);
+        } else {
+            toggleButton.setEnabled(false);
+        }
+    }
+
+
+    @Override
+    public void onLoaderReset(Loader loader) {
```
shobhitagarwal1612: Same as above

share_app/src/main/java/org/odk/share/adapters/InstanceAdapter.java
```diff
@@ -0,0 +1,85 @@
...
+        long id = cursor.getLong(cursor.getColumnIndex(InstanceProviderAPI.InstanceColumns._ID));
+        holder.checkBox.setChecked(selectedInstances.contains(id));
+    }
+
+    @Override
+    public int getItemCount() {
+        return cursor.getCount();
+    }
+
+    public class InstanceHolder extends RecyclerView.ViewHolder {
```
shobhitagarwal1612: Class declaration should be at the bottom. Just try  IntelliJ's auto-reformat `Ctrl` + 'Shift` + 'f'

shobhitagarwal1612: Also, the scope should be package-private as this holder is not going to be used outside of this class

PR #18, opened 2018-05-23
Activity to show Wifi scan results

lakshyagupta21:
This PR includes
* `WRITE_SETTINGS` permission to a manifest which is needed whenever new hotspot configuration is added.
* Icons added for strengths of both open and close networks
* Wifi network scanning (checks for suffix `-odk-share` which is created using same app)
* Wifi Turn on/off cases handled
* Dialog added for password protected networks
* Changed popup menu text color and overflow icon color to white

ezgif com-resize


lakshyagupta21:
> created using this app

I was thinking about this, but to check we have two options either hardcode the ssid or concatenate some string like "odk-share" at the end of ssid so that it can be verified that it is created using this app.

share_app/src/main/AndroidManifest.xml
```diff
@@ -5,6 +5,7 @@
     <uses-permission android:name="android.permission.CHANGE_WIFI_STATE" />
     <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
+    <uses-permission android:name="android.permission.WRITE_SETTINGS" />
```
shobhitagarwal1612: Why is this permission required?

lakshyagupta21: Whenever we create new WifiConfiguration it requires this permission.

share_app/src/main/java/org/odk/share/controller/Wifi.java
```diff
@@ -0,0 +1,29 @@
...
+
+import android.content.Context;
+import android.net.wifi.ScanResult;
+import android.net.wifi.WifiManager;
+
+/**
+ * Created by laksh on 5/22/2018.
+ */
+
+public class Wifi {
```
shobhitagarwal1612: The name of the class does not reflect what it is doing. A better name could be `WifiHelper` or something else

share_app/src/main/java/org/odk/share/controller/Wifi.java
```diff
@@ -0,0 +1,29 @@
...
+
+    private WifiManager wifiManager;
```
shobhitagarwal1612: Can be declared as `final`

share_app/src/main/java/org/odk/share/controller/WifiHotspot.java
```diff
@@ -24,9 +26,11 @@
     private WifiManager wifiManager;
     private WifiConfiguration lastConfig;
     private WifiConfiguration currConfig;
-    private static final String ssid = "ODK-Share";
+    private Context context;
+    private static final String ssid = "hotspot";
 
     public WifiHotspot(Context context) {
```
shobhitagarwal1612: Better name of this class would be something like WifiHotspotManager or WifiHotspotHelper

share_app/src/main/java/org/odk/share/adapters/WifiResultAdapter.java
```diff
@@ -0,0 +1,100 @@
...
+        return new WifiHolder(view);
+    }
+
+    @Override
+    public void onBindViewHolder(@NonNull final WifiHolder holder, int position) {
+        holder.title.setText(wifiScanResult.get(holder.getAdapterPosition()).SSID);
+        holder.itemView.setOnClickListener(v -> listener.onItemClick(v, holder.getAdapterPosition()));
+
+        int level = WifiManager.calculateSignalLevel(wifiScanResult.get(holder.getAdapterPosition()).level, 4);
+        if (Wifi.isClose(wifiScanResult.get(holder.getAdapterPosition()))) {
```
shobhitagarwal1612: The below logic should be either in the holder class or the code can be refactored to create a custom `WifiView` which basically accepts an integer value and updates it's icon

PR #20, opened 2018-05-26
Enable hotspot from settings

### LINKED ISSUE #19: Problem connecting to created wifi hotspot network
Issue noticed in Android versions >= 7, below 7 other devices are able to connect to created hotspot network.

lakshyagupta21:
Closes #19 

Devices with android version >= 7 is able to create and enable hotspot network from the app but no other devices are able to connect to it.
To fix this issue, a check is added so that user can navigate to hotspot settings and then have to enable from their, then only other devices will be able to connect to it.

### DIFF
--- a/share_app/src/main/AndroidManifest.xml
+++ b/share_app/src/main/AndroidManifest.xml
@@ -25,7 +25,6 @@
         </activity>
 
         <activity android:name=".activities.WifiActivity"/>
-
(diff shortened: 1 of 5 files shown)


lakshyagupta21:
Should I add a dialog box to give an idea of what and why it is needed to enable the hotspot before navigating to it?

shobhitagarwal1612:
Sure.

PR #23, opened 2018-06-02
Tasks to send and receive form instances using socket connection

### LINKED ISSUE #11: Establish socket connection over wifi-hotspot

lakshyagupta21:
Closes #11 

I've created a separate activity to trigger hotspot creation and connection and after a connection is established it will start sending the forms and will finish once receiver receives the form.
To start receiving the form receiver need to tap on the wifi network then it will start receiving the form, once the tasks get finished, it will return back to MainActivity.

There are few things that I still need to take care like
- [x] Dialog show/dismiss
- [x] Checking for available socket ports
- [x] Showing Error messages, success messages
- [ ] Storage (In ODK directory right now forms are saved in Share folder)
- [x] Receiving and Sending of forms (checking whether form exists or not)
- [ ] Sending the details of instances (right now only few are sent)
- [ ] Handling the `cancel` task
- [ ] Storing the form and instances detail in Database.

ezgif com-gif-maker 2

### DIFF
--- a/share_app/src/main/java/org/odk/share/activities/WifiActivity.java
+++ b/share_app/src/main/java/org/odk/share/activities/WifiActivity.java
@@ -64,6 +64,8 @@ public class WifiActivity extends AppCompatActivity implements ProgressListener 
     private WifiHelper wifiHelper;
     private String wifiNetworkSSID;
 
+    private String alertMsg;
+
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
@@ -124,6 +126,7 @@ public class WifiActivity extends AppCompatActivity implements ProgressListener 
             showPasswordDialog(scanResultList.get(i));
         } else {
             // connect
+            alertMsg = getString(R.string.connecting_wifi);
             showDialog(DIALOG_CONNECTING);
             wifiNetworkSSID = scanResultList.get(i).SSID;
             wifiHelper.connectToWifi(scanResultList.get(i), null);
@@ -304,7 +307,7 @@ public class WifiActivity extends AppCompatActivity implements ProgressListener 
         switch (id) {
             case DIALOG_DOWNLOAD_PROGRESS:
                 progressDialog = new ProgressDialog(this);
-                progressDialog.setMessage("Receiving");
+                progressDialog.setMessage(alertMsg);
                 progressDialog.setIndeterminate(true);
                 progressDialog.setCancelable(false);
 
@@ -323,7 +326,7 @@ public class WifiActivity extends AppCompatActivity implements ProgressListener 
                 return progressDialog;
             case DIALOG_CONNECTING:
                 progressDialog = new ProgressDialog(this);
-                progressDialog.setMessage("Connecting to wifi");
+                progressDialog.setMessage(alertMsg);
                 progressDialog.setIndeterminate(true);
                 progressDialog.setCancelable(false);
                 progressDialog.show();
@@ -347,11 +350,13 @@ public class WifiActivity extends AppCompatActivity implements ProgressListener 
             Timber.e(e);
         }
         Toast.makeText(this, "Files Received", Toast.LENGTH_LONG).show();
-        finish();
+        createAlertDialog("Transfer result", result);
     }
 
     @Override
     public void progressUpdate(int progress, int total) {
+        alertMsg = "Receiving form " + progress + " of " + total;
+        progressDialog.setMessage(alertMsg);
     }
 
     @Override
@@ -362,5 +367,24 @@ public class WifiActivity extends AppCompatActivity implements ProgressListener 
         } catch (Exception e) {
             Timber.e(e);
         }
+    }
+
+    private void createAlertDialog(String title, String message) {
+        alertDialog = new AlertDialog.Builder(this).create();
+        alertDialog.setTitle(title);
+        alertDialog.setMessage(message);
+        DialogInterface.OnClickListener quitListener = new DialogInterface.OnClickListener() {
+            @Override
+            public void onClick(DialogInterface dialog, int i) {
+                switch (i) {
+                    case DialogInterface.BUTTON_POSITIVE:
+                        finish();
+                        break;
+                }
+            }
+        };
+        alertDialog.setCancelable(false);
+        alertDialog.setButton(DialogInterface.BUTTON_POSITIVE, getString(R.string.ok), quitListener);
+        alertDialog.show();
     }
 }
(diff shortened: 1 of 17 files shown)


shobhitagarwal1612:
I tried to send a form instance but it didn't work. Also, since the current state of the connection is not shown anywhere, so it's hard to understand what's wrong.

shobhitagarwal1612:
So a couple of ideas to make the testing/debugging a bit easier.

1. Add the device manufacturer's name to the created wifi hotspot eg `ODK-Share-Xiaomi`

2. Display current connectivity status, CONNECTED, IDLE

3. Create wifi-hotspot on the receiving end. This will ensure the sender that if the `ODK-Share-device` hotspot is available then it is ready and the socket connection can be established to send the forms. 

4. Display the current progress on both the ends.

shobhitagarwal1612:
Another idea is to get the intention of the user as soon as the app is opened i.e. whether the user wants to send, receive the forms or review received forms.

Send forms  > Choose form instances > Select a wifi > Start socket connection > Start sending

Receive forms > Start wifi hotspot > Wait for the connection > Display a UI for the current transfer progress

Review forms > Show a last received instance list.

@lognaturel @lakshyagupta21 What do you think?

lakshyagupta21:
> whether the user wants to send, receive the forms or review received forms.

I think this is the best way to send, receive and review forms. 
I have created wifi hotspot at the sender end because in future there may be some requirement in which enumerator want to send forms to multiple enumerators that case receivers can connect to it and sender can accept as many connections as it can connect to. 

What do you say?

lakshyagupta21:
> Also, since the current state of the connection is not shown anywhere, so it's hard to understand what's wrong.

@shobhitagarwal1612 Can you share the logs, that will  help me debug things, if that is related to `port already in use` error then I'll resolve this in my next commit.

shobhitagarwal1612:
> Can you share the logs, that will help me debug things, if that is related to port already in use error then I'll resolve this in my next commit.

There were no error logs on the receiver end. I clicked on the wifi hotspot from within the `View Wifi` screen, and a few moments later a toast appeared with `file received` message. But I couldn't find anything downloaded to the device storage.

What is the destination folder for the downloaded instances?

lakshyagupta21:
> What is the destination folder for the downloaded instances?

Downloads directory

lakshyagupta21:
> I've discovered that if I put the Tab A on the "View Wifi" screen I sometimes get a "Receiving" dialog but that also just gets stuck and there's no way to cancel so I have to close the app.

It would be very helpful if you share the form details like its XML and media resources and the entry corresponding to it in DB.

I will add status to the dialog and button to cancel it in next commits.

> On the Moto G4 I can successfully create the hotspot. Once it's up, the Galaxy Tab A does connect. I sometimes get this crash on the Tab A:

Can you share the SSID of Wifi Network that you're trying to connect?

lognaturel:
The form I tried to send from both devices is [this one](https://github.com/opendatakit/collect/files/2028746/sms_form.xml.txt). Could it be a problem that the same one was on both devices? Will need to pull the db a bit later. No media.

> SSID of Wifi Network

`ODK-Share-odk-share`

shobhitagarwal1612:
@lakshyagupta21 The file transfer worked successfully but I did notice a crash if I connect the receiving device to the hotspot before sending the forms from the sender.

```

File: share_app/src/main/java/org/odk/share/controller/WifiHelper.java [hunk 1]
```diff
@@ -11,11 +18,77 @@
...
+    public static WifiHelper wifiHelper;
+    private Context context;
 
-    public WifiHelper(Context context) {
+    private WifiHelper(Context context) {
+        this.context = context;
         wifiManager = (WifiManager) context.getApplicationContext().getSystemService(Context.WIFI_SERVICE);
     }
 
+    public static synchronized WifiHelper getInstance(Context context) {
```
shobhitagarwal1612: Creating a singleton is not a good practice and later leads to complications. Is there any specific use case because of which you had to do it? If not then please try not to use it

share_app/src/main/java/org/odk/share/activities/WifiActivity.java
```diff
@@ -307,4 +301,89 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+                progressDialog.setIndeterminate(true);
+                progressDialog.setCancelable(false);
+                progressDialog.show();
+                return progressDialog;
+            default:
+                return null;
+        }
+    }
+
+    public static String getAccessPointIpAddress(Context context) {
```
shobhitagarwal1612: This method can be moved to WifiHelper since it is a helper method which is not coupled with it's usage

lakshyagupta21:
@lognaturel Wifi Hotspot button on MainActivity created some confusion I think due to which you and @shobhitagarwal1612 noticed the crash, that button was added just to test few things related to the hotspot. But now I have added the hotspot implementation in `SendActivity`, so I removed it from the main screen to avoid the confusion created.

Try it again and if you still face the crash let me know.

shobhitagarwal1612:
@lakshyagupta21 File transfer is working perfectly. I was able to create a hotspot on Android 7.0 and send the forms/instances instantly. The only problem is that on Android 8.0, the hotspot is not getting renamed. So because of this, I am unable to receive the data. But other than that, everything is working fine.

So let's address those issues separately in another PR. 
Another suggestion is, let's not append the port number to the hotspot name. Here is what we can do to make the flow a bit smoother.
 - After choosing the forms and starting the hotspot, display a status "Waiting for the enumerator..."
 - Along with that display a QR code containing the info of the current hotspot network 
 - On the receiving end, show a list of nearby connections and also a button to connect using QR code
 - As soon as the connection is established update the UI on both the ends showing current progress of the file transfer

Also, using FTP for file transfer is also pretty fast and simple. So later we can look into that option as well.

@lognaturel @lakshyagupta21 What are your thoughts?

lakshyagupta21:
> The only problem is that on Android 8.0, the hotspot is not getting renamed.

There are some issues going on with Android 8.0 related to Wifi Hotspot, so I think one way to resolve this is to create hotspot with the already existing configuration (without changing anything like ssid and password).




> Along with that display a QR code containing the info of the current hotspot network

So what do you suggest showing QR Code in dialog and this way we're forcing users to connect the devices by scanning the QR code.




> As soon as the connection is established update the UI on both the ends showing current progress of the file transfer

Currently it shows the progress at the receiving end only but will update the sender end so that it shows both "Waiting" status as well as transfer progress

We can remove the port number from SSID and add QR Code implementation in another PR there are already lot of things in this PR, it will be easier to take these things in another PR.

shobhitagarwal1612:
>  it will be easier to take these things in another PR.

Absolutely, you can create small issues out of these points and solve them one at a time in different PRs

lakshyagupta21:
@shobhitagarwal1612 I will file issues of each subtask separately and submit different PR.

Please review and merge if everything is fine.

shobhitagarwal1612:
Just a couple of string resources to be converted and this PR looks good to be merged.

PR #25, opened 2018-06-09
Scan QR Code to connect

### LINKED ISSUE #21: Connect hotspot by scanning QR Code
1. Add functionality to generate QR Code for hotspot - QR Code can contain two or three attributes depending on whether wifi is protected or not. Three attributes -> `ssid`, `isProtected`, `password`
2. Add functionality to scan the QR code in `WifiActivity` so that hotspot can be connected.

lakshyagupta21:
Closes #21 

Scan QR Code to connect to Wifi Hotspot. 
Port number is removed from the ssid's, the only way to connect to wifi hotspot is by scanning the QR Code which is generated at the sender's end. This way port number and password's can be shared to the receiver.

qrcode

### DIFF
--- a/share_app/build.gradle
+++ b/share_app/build.gradle
@@ -40,4 +40,8 @@ dependencies {
 
     implementation "com.google.zxing:core:3.3.0"
     implementation "com.journeyapps:zxing-android-embedded:3.5.0"
+
+    // RxJava 2:
+    implementation "io.reactivex.rxjava2:rxandroid:2.0.2"
+    implementation "io.reactivex.rxjava2:rxjava:2.1.10"
 }
--- a/share_app/build.gradle
+++ b/share_app/build.gradle
@@ -37,4 +37,7 @@ dependencies {
     annotationProcessor "com.jakewharton:butterknife-compiler:${rootProject.butterknifeVersion}"
 
     implementation "com.jakewharton.timber:timber:4.7.0"
+
+    implementation "com.google.zxing:core:3.3.0"
+    implementation "com.journeyapps:zxing-android-embedded:3.5.0"
 }
--- a/share_app/src/main/AndroidManifest.xml
+++ b/share_app/src/main/AndroidManifest.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
     package="org.odk.share">
 
     <uses-permission android:name="android.permission.CHANGE_WIFI_STATE" />
@@ -28,10 +29,14 @@
         <activity android:name=".activities.WifiActivity" />
         <activity android:name=".activities.InstancesList" />
         <activity android:name=".preferences.SettingsPreference" />
-
+        <activity android:name=".activities.SendActivity"/>
         <service android:name=".services.HotspotService" />
 
-        <activity android:name=".activities.SendActivity"></activity>
+        <activity
+            android:name="com.journeyapps.barcodescanner.CaptureActivity"
+            android:screenOrientation="portrait"
+            tools:replace="android:screenOrientation"
+            android:stateNotNeeded="true"/>
     </application>
 
 </manifest>
--- a/share_app/src/main/java/org/odk/share/activities/SendActivity.java
+++ b/share_app/src/main/java/org/odk/share/activities/SendActivity.java
@@ -8,10 +8,14 @@ import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
 import android.content.IntentFilter;
+import android.net.wifi.WifiConfiguration;
+import android.os.Bundle;
 import android.support.v4.content.LocalBroadcastManager;
 import android.support.v7.app.AlertDialog;
 import android.support.v7.app.AppCompatActivity;
-import android.os.Bundle;
+import android.support.v7.widget.Toolbar;
+import android.widget.ImageView;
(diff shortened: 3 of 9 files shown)


shobhitagarwal1612:
This looks great @lakshyagupta21 and the code looks neat :) Great work!!

> the only way to connect to wifi hotspot is by scanning the QR Code which is generated at the sender's end. 

I think it is important to have the functionality of being able to manually enter the `ip:port` as well. 

So for sender's side: 
The label `waiting for connection to accept` can be moved to the top. Below the QR code we can write something like "For connection, scan this QR code or type "<ip adress>:<port>" after connecting to WiFi Hotspot <hotspot name>

For receiver's side:
Clicking on WiFi hotspot name will only connect to that network and if successful, a dialog will appear in which the relevant information can be entered.

@lakshyagupta21 @lognaturel what are your thoughts?

lakshyagupta21:
> Clicking on WiFi hotspot name will only connect to that network and if successful, a dialog will appear in which the relevant information can be entered.

After successful connection, we can have a dialog box asking the user to enter the details or will give an option to scan the QR code to get the details.

shobhitagarwal1612:
If the user has access to scan QR code then pre connecting to that QR code seems redundant as the QR code can also contain info regarding which network to connect to

lognaturel:
> Would it be better if we don't include the text "port" in the dialog?

I think it's ok. Users will mostly pattern match anyway but for those who do know what it means it's provides a little context.

This worked perfectly the first time I tried it and then it failed several times in a row. I am using my Moto G4 as the sender and the Galaxy Tab A as the receiver. After that one successful transfer, on every trial the Tab A (receiver) hung at "Connecting to wifi" after I input the port or scanned the QR code. I tried everything I can think of -- killing the app on both sides, filling a new form in Collect. Then all of a sudden it started working again and now it seems to work consistently. Any ideas what may have caused that problem? Is there any additional info I can provide to help troubleshoot?

Small crash that can be filed separately -- if I rotate the screen while it's hanging I get
```
java.lang.RuntimeException: Unable to destroy activity {org.odk.share/org.odk.share.activities.WifiActivity}: java.lang.NullPointerException: Attempt to invoke interface method 'java.util.Iterator java.util.List.iterator()' on a null object reference
	at android.app.ActivityThread.performDestroyActivity(ActivityThread.java:4896)
	at android.app.ActivityThread.handleDestroyActivity(ActivityThread.java:4914)
	at android.app.ActivityThread.handleRelaunchActivity(ActivityThread.java:5173)
	at android.app.ActivityThread.access$1100(ActivityThread.java:219)
	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1741)
	at android.os.Handler.dispatchMessage(Handler.java:102)
	at android.os.Looper.loop(Looper.java:145)
	at android.app.ActivityThread.main(ActivityThread.java:6939)
	at java.lang.reflect.Method.invoke(Native Method)
	at java.lang.reflect.Method.invoke(Method.java:372)
	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1404)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1199)
```

When I try to use the Galaxy Tab A as a sender, I go directly to the "Send forms" screen without any instructions to turn on the hotspot. If I manually go to Settings and Mobile hotspot and tethering, then try to toggle the mobile hotspot on, I get the message "Insert a SIM card to use Mobile Hotspot and other tethering functions." 🤯

Does anyone have any other ideas for turning on the hotspot? This is the most popular device Collect runs on so it is important to get some way for it to work. I was able to force the hotspot on by launching the Wifi Hotspot app but Share doesn't then configure the hotspot.

lakshyagupta21:
> Is there any additional info I can provide to help troubleshoot?

Can you provide logs of both sender and receiver end?



> I get the message "Insert a SIM card to use Mobile Hotspot and other tethering functions."

It looks like this issue is with only Samsung devices. While creating hotspot I'm just appending `-odk-share` to the ssid so that at the receiver it can be known that the wifi is created by share app, if possible I'll try to use the existing configuration so that more devices can be covered.

lognaturel:
> Can you provide logs of both sender and receiver end?

Which logs would be helpful, exactly?

> try to use the existing configuration

The fundamental problem is that it doesn't even seem possible to enable a hotspot without downloading a secondary app. I hope I'm wrong about this but I can't currently find another way.

> It looks like this issue is with only Samsung devices. 

Samsung is by far the most popular manufacturer so we do need to address this somehow.

lakshyagupta21:
> Which logs would be helpful, exactly?

Even I noticed this behaviour, sometimes it works sometimes not, the issue is that HotspotSendTask, not calls `doInBackground` method, I'm not sure why this is happening but still, I want to know about the devices you're using, you can copy and paste all the logs from the logcat window.

lognaturel:
Sounds good, @lakshyagupta21. I'll try to remember to keep logcat running when I do trials and to copy the output for you.

I think this can be merged and issues can be filed for the things we've identified to look into. ✨What do you think, @shobhitagarwal1612?

shobhitagarwal1612:
Me and Lakshya had some discussion regarding this on slack. Here is some conversations pasted from there.
"So the problem does not occur if I successfully start the client and receive the forms. The problem occurs only if I press back button after selecting the forms and starting wifi hotspot without letting the client connect to the network. So what happens is, the old async task is not destroyed and a new one is created. That's why for the new task, the doInBackground is not reaching as the old one is still there and the socket connection is occupied."

> I think this can be merged and issues can be filed for the things we've identified to look into

Yes, agreed.

PR #28, opened 2018-06-14
Database created and added row corresponding to each form and instance

### LINKED ISSUE #26: Create database and save information received from the receiver end
* Create share db for storing send/receive data
* Store instance information in instance db
* Store instance information share db

lakshyagupta21:
Closes #26 

Added row in a `forms.db` and `instances.db` 
Added row in `share.db` at sender as well as receiver end.

For now, I've added forms and instances under share directory and added a path of share folder in collect databases for each instance and form received.

**Note**: Before testing this PR make sure you copy odk/metadata and save it somewhere so that your original table doesn't get updated.


lakshyagupta21:
@shobhitagarwal1612 Now its ready, Please review

shobhitagarwal1612:
Transferred 2 instances of All-Widgets. Below is the data in the dbs of the receiver.

Instances db:
image

Forms db:
image

Share db:
image

shobhitagarwal1612:
It looks fine to me.
@lognaturel  What do you think?

share_app/src/main/java/org/odk/share/database/ShareDatabaseHelper.java
```diff
@@ -0,0 +1,76 @@
...
+    private static final int DATABASE_VERSION = 1;
+
+    public ShareDatabaseHelper() {
+        super(new DatabaseContext(Share.METADATA_PATH), DATABASE_NAME, null, DATABASE_VERSION);
+        Timber.d("Context " + Share.METADATA_PATH);
+    }
+
+    @Override
+    public void onCreate(SQLiteDatabase db) {
+        Timber.d("Database creation started");
```
shobhitagarwal1612: These logs can be dropped

share_app/src/main/java/org/odk/share/database/ShareDatabaseHelper.java
```diff
@@ -0,0 +1,76 @@
...
+        createInstancesTable(db);
+        Timber.d("Database creation stopped");
```
shobhitagarwal1612: This log can be dropped

share_app/src/main/java/org/odk/share/database/ShareDatabaseHelper.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    @Override
+    public void onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
+
+        Timber.d("onUpgrade");
```
shobhitagarwal1612: It would be even better if the `oldVersion` and `newVersion` are logged as well

share_app/src/main/java/org/odk/share/tasks/DownloadJob.java
```diff
@@ -178,42 +192,77 @@ private void readForm() {
...
+                values.put(INSTANCE_FILE_PATH, path);
+                values.put(JR_FORM_ID, formId);
+                values.put(JR_VERSION, formVersion);
+                Uri uri = new InstancesDao().saveInstance(values);
+
+                // Add row in share table
+                ContentValues shareValues = new ContentValues();
+                shareValues.put(INSTANCE_ID, Long.parseLong(uri.getLastPathSegment()));
+                shareValues.put(TRANSFER_STATUS, STATUS_FORM_RECEIVE);
+                new ShareDatabaseHelper().insertInstance(shareValues);
```
shobhitagarwal1612: The above two values can be sent as a parameter to the `insertInstance` method

share_app/src/main/java/org/odk/share/tasks/UploadJob.java
```diff
@@ -310,11 +312,31 @@ private void sendInstances(List<String> instanceIds, int progress, int total) {
...
-                    Timber.d("Progress " + progress + " " + total);
+
                     sendInstance(instance);
+
+                    // add row in share table
+                    ContentValues values = new ContentValues();
+                    values.put(INSTANCE_ID,
+                            c.getLong(c.getColumnIndex(InstanceProviderAPI.InstanceColumns._ID)));
+                    values.put(TRANSFER_STATUS, STATUS_FORM_SENT);
+                    new ShareDatabaseHelper().insertInstance(values);
```
shobhitagarwal1612: The above two values can be sent as a parameter to the `insertInstance` method.

share_app/src/main/java/org/odk/share/tasks/UploadJob.java
```diff
@@ -310,11 +312,31 @@ private void sendInstances(List<String> instanceIds, int progress, int total) {
...
+
+                    dos.writeUTF(displayName);
+
+                    if (submissionUri == null) {
+                        dos.writeUTF("-1");
+                    } else {
+                        dos.writeUTF(submissionUri);
+                    }
+
+                    rxEventBus.post(new UploadEvent(UploadEvent.Status.UPLOADING, ++progress, total));
```
shobhitagarwal1612: Progress should be updated after the uploading is done. So after `sendInstance`, right?

share_app/src/main/java/org/odk/share/database/ShareDatabaseHelper.java
```diff
@@ -0,0 +1,76 @@
...
+public class ShareDatabaseHelper extends SQLiteOpenHelper {
+
+    private static final String DATABASE_NAME = "share.db";
+    private static final String SHARE_TABLE_NAME = "share";
+
+    private static final int DATABASE_VERSION = 1;
+
+    public ShareDatabaseHelper() {
+        super(new DatabaseContext(Share.METADATA_PATH), DATABASE_NAME, null, DATABASE_VERSION);
+        Timber.d("Context " + Share.METADATA_PATH);
```
shobhitagarwal1612: This log can be dropped

share_app/src/main/java/org/odk/share/database/DatabaseContext.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Created by laksh on 6/13/2018.
+ */
+
+public class DatabaseContext extends ContextWrapper {
+    private String path;
+
+    public DatabaseContext(String path) {
+        super(Share.getInstance());
```
shobhitagarwal1612: Application context should be passed as a parameter instead of being accessed via singleton

share_app/src/main/java/org/odk/share/database/ShareDatabaseHelper.java
```diff
@@ -0,0 +1,76 @@
...
+import static org.odk.share.dto.TransferInstance.TRANSFER_STATUS;
+
+/**
+ * Created by laksh on 6/13/2018.
+ */
+
+public class ShareDatabaseHelper extends SQLiteOpenHelper {
+
+    private static final String DATABASE_NAME = "share.db";
+    private static final String SHARE_TABLE_NAME = "share";
```
lognaturel: Are you planning to store status info about instances and forms in the same table or different ones? Is `share` the best table name?

lognaturel:
Looks good overall, just question about table name that I've put inline. 🤘

share_app/src/main/java/org/odk/share/database/ShareDatabaseHelper.java
lakshyagupta21: I'll be storing all instances transfer and feedback details in the same table, I'm not good at naming things, that's why I named it as `share`. Can you suggest some other name?

lognaturel: 😊Naming is hard. Maybe something like `transfer-metadata` or `transfer-summary` or even just `transfers`?

lognaturel: I just realized that you named your DTO well so that will bring things nicely in line. 👍 https://twitter.com/timbray/status/817025379109990402?cn=cmVwbHk%3D 😄

share_app/src/main/java/org/odk/share/tasks/DownloadJob.java
lakshyagupta21: I wrote like this because sometimes I may get a feedback from the sender and sometimes not. So I don't know when I'm gonna receive it, maybe in future, some more fields may get added to a database that's why I didn't send all values as a parameter. 
What do you think?

share_app/src/main/java/org/odk/share/tasks/DownloadJob.java
```diff
@@ -178,42 +192,77 @@ private void readForm() {
...
+                values.put(CAN_EDIT_WHEN_COMPLETE, "true");
+                values.put(SUBMISSION_URI, submissionUri);
+                values.put(INSTANCE_FILE_PATH, path);
+                values.put(JR_FORM_ID, formId);
+                values.put(JR_VERSION, formVersion);
+                Uri uri = new InstancesDao().saveInstance(values);
+
+                // Add row in share table
+                ContentValues shareValues = new ContentValues();
+                shareValues.put(INSTANCE_ID, Long.parseLong(uri.getLastPathSegment()));
```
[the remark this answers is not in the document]
lakshyagupta21: I'm storing info about the instances, and for forms transfer, I'm not keeping any track of `forms`, if the forms are received then I'm just updating the `forms` table. Is it required to track the forms also?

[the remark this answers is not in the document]
lakshyagupta21: > Why are you calling getLastPathSegment()

I'm just storing the instanceID, which was added to a `instances` table.
`getLastPathSegment()` extracts the id of the newly added row. I'm not storing the `uri`.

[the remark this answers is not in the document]
lognaturel: Got it, that makes sense. Since instances are specific records it's important to know who has had access to them when. A form with a particular form_id can be duplicated a bunch of times and it doesn't matter so much where they come from. So that works.

What about the name `INSTANCE_ID`?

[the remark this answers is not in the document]
lognaturel: > getLastPathSegment() extracts the id of the newly added row

Great, that makes sense. I haven't had a chance to try this myself because I don't have two devices at the moment. @shobhitagarwal1612's screenshot confused me because all the instanceID values are showing up as single digits.

share_app/src/main/java/org/odk/share/tasks/DownloadJob.java
lakshyagupta21: We will be updating the table after review as well as after every instance transfer.  There are two cases one when instance is received other one is any user marks instance as reviewed after giving the feedback. But sometime user doesn't give any feedback so in that case we have to write multiple methods for each case.

shobhitagarwal1612:
@lognaturel Please merge the PR if the changes look good to you

PR #30, opened 2018-06-14
Introduce RxJava2, Dagger and create Sender/Receiver Services for uploading/download

### LINKED ISSUE #31: Handle cancel event of Async Tasks
Noticed in PR #25 
Not able to send/receive forms if a user tries again after an activity is destroyed by pressing back button or cancel button of a progress bar.

New Async task won't start until the previous one gets finished. So need to handle the cancelling event of Async tasks.

shobhitagarwal1612:
Closes #31 

@lakshyagupta21 @lognaturel I have added support for Dagger 2 (for dependency injection) and RxJava2 (for writing async event-based code)

The purpose of this PR is to add necessary dependencies and a demo of how we can use it to make the code event based using the observable pattern. 
Please have a look and share your thoughts


shobhitagarwal1612:
In the latest commit, I have used the Evernote's job creator (as used in Collect) to create a queue of download jobs that get triggered asynchronously.

I have a plan to implement a similar approach for the uploading task also. Please give your feedbacks on the same too.

lakshyagupta21:
Thanks, @shobhitagarwal1612 !!

lakshyagupta21:
There are some lint issues which will be fixed in next PR.

PR #34, opened 2018-06-21
Minor improvements

shobhitagarwal1612:
- Cancel all pending jobs from the last session (if any)
 - Fix crash due to NPE by adding a null check
 - QR code now fills the entire width for better visibility


shobhitagarwal1612:
@lakshyagupta21 Please have a look at this PR.

PR #35, opened 2018-06-23
Home screen UI

### LINKED ISSUE #32: Main Screen to show all the forms available
1. Show all the forms available in device.
2. Tapping on any category will open a activity which will show instances of that category only.
3. Add send and receive buttons to send the forms (Activity will show all the instances of all the forms)

lakshyagupta21:
Send and receive buttons covers more than the expected space. It contains circular view and then under that, we have a text showing button label, I think it doesn't look good when the forms list covers the whole screen, so adding simple buttons at the bottom would be easier for users to use and understand as it looks somewhat similar to collect UI.

- [x] Show all the forms available in device.
- [x] Add send and receive buttons to send the forms (Activity will show all the instances of all the forms)

ezgif com-resize


shobhitagarwal1612:
@lakshyagupta21 This PR fixes only the first point of the issue #32. So please remove the `Closes #32` to references #32. Also, in the issue, the points should be converted to a checklist.

shobhitagarwal1612:
Also, please have a look at this [commit](https://github.com/opendatakit/share/commit/5e2afb3f05b6c489e1f7ca4e5ed1d47e2c00db8c)

Since data in most of the screens would be populated from the database, so I thought it would be better if there is a common implementation of the adapter which all of the activities can implement. Another advantage of using this implementation is, since this contains `data observers` so if there is any change in the dataset of the cursor, adapter will automatically update the UI

shobhitagarwal1612:
Likewise, you won't have to apply all of the checks in each of the implemented adapter

PR #38, opened 2018-06-27
Tab Activity for forms

### LINKED ISSUE #37: Screen for showing stats, sent, received, review forms
Create screen with tabs that show  statistics, sent forms, received forms, reviewed forms, etc.
I've added 2-3 wireframes showing different options of selection i.e, Selection on long tap and checkbox
I wanted to know which option would be preferable.

lakshyagupta21:
Closes #37 

Tabs for showing activity of forms added.

ezgif com-resize


share_app/src/main/java/org/odk/share/activities/MainActivity.java
```diff
@@ -118,5 +120,9 @@ public void onLoaderReset(@NonNull Loader loader) {
     private void onItemClick(View view, int position) {
         Cursor cursor = formAdapter.getCursor();
         cursor.moveToPosition(position);
+        Intent intent  = new Intent(this, InstanceManagerTabs.class);
+        intent.putExtra(FORM_VERSION, cursor.getString(cursor.getColumnIndex(FormsProviderAPI.FormsColumns.JR_VERSION)));
```
shobhitagarwal1612: We already parse form id and version for the item. So let's reuse that instead of using cursor to fetch it

share_app/src/main/java/org/odk/share/dto/TransferInstance.java
```diff
@@ -30,11 +39,11 @@ public void setId(Long id) {
         this.id = id;
     }
 
-    public Boolean getReviewed() {
+    public Integer getReviewed() {
```
shobhitagarwal1612: Use int instead of wrapper class

share_app/src/main/java/org/odk/share/dto/TransferInstance.java
```diff
@@ -30,11 +39,11 @@ public void setId(Long id) {
...
         return isReviewed;
     }
 
-    public void setReviewed(Boolean reviewed) {
+    public void setReviewed(Integer reviewed) {
```
shobhitagarwal1612: Same as above

lognaturel:
This is looking very nice to me, @lakshyagupta21! Is there any specific feedback you're looking for? It looks like you're still planning to add to this branch?

lakshyagupta21:
@lognaturel I still have to add few action buttons in each tabs. 

If there are several versions of the same forms available in a device and for each version, we also have some instances stored in our device which may be filled in the same device or received by some other device then in that case what we should do? Currently, all the available versions of forms are shown on the Home screen of the app, I think we don't do that in collect, we only show the latest version.

share_app/src/main/java/org/odk/share/activities/MainActivity.java
lakshyagupta21: How to reuse that? We can access only the `TextView` of the Item right?

lakshyagupta21:
tab

Buttons from the `Received` tabs are removed so that user can view the form and review by tapping on it.

PR #40, opened 2018-07-08
Hotspot connection issue resolved for devices >= Oreo

### LINKED ISSUE #33: Unable to set the wifi hotspot name in android oreo
In android oreo, only system apps which are signed and have the below permission can change the configuration.
```
<permission
        android:name="android.permission.OVERRIDE_WIFI_CONFIG"
        android:protectionLevel="signature|privileged" />
```

https://stackoverflow.com/questions/46392277/changing-android-hotspot-settings

So, this causes a problem if the sender is unable to change the name and uses that name to create the QR Code which later fails due to mismatch.
Since there is no way to change the name, I suggest we provide a setting to configure hotspot name in setting for devices < Oreo and read the actual wifi hotspot name after turning it on to create the QR code or writing instructions

lakshyagupta21:
Fixes #33 

[WifiManager](https://developer.android.com/reference/android/net/wifi/WifiManager#startLocalOnlyHotspot(android.net.wifi.WifiManager.LocalOnlyHotspotCallback,%20android.os.Handler))
For devices having android version >= 8 they have added public API which can be used to enable and disable the hotspot. 3rd party apps can only enable/disable it, they are not allowed to change the WifiConfiguration. 

For Oreo or greater versions, user have to scan the QR Code to connect to the network.

Useful Links- 
[https://stackoverflow.com/questions/45984345/how-to-turn-on-off-wifi-hotspot-programmatically-in-android-8-0-oreo/45996578#45996578](https://stackoverflow.com/questions/47700717/change-wifi-hotspots-ssid-and-password-in-android-oreo-8-x)
[https://stackoverflow.com/questions/45984345/how-to-turn-on-off-wifi-hotspot-programmatically-in-android-8-0-oreo/45996578#45996578](https://stackoverflow.com/questions/45984345/how-to-turn-on-off-wifi-hotspot-programmatically-in-android-8-0-oreo/45996578#45996578)

### DIFF
--- a/share_app/src/main/res/values/strings.xml
+++ b/share_app/src/main/res/values/strings.xml
@@ -5,7 +5,9 @@
     <string name="stop">Stop</string>
     <string name="hotspot_settings_dialog">Enable portable hotspot from the settings.</string>
     <string name="hotspot_name_suffix">-odk-share</string>
+    <string name="hotspot_name_prefix_oreo">AndroidShare</string>
     <string name="hotspot_already_running">Hotspot already running</string>
+    <string name="scan_alert_oreo">Scan QR Code to connect to this network</string>
 
     <string name="saved_forms">Saved Forms</string>
     <string name="send_selected">Send</string>
--- a/share_app/src/main/AndroidManifest.xml
+++ b/share_app/src/main/AndroidManifest.xml
@@ -11,6 +11,8 @@
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
+
+    <!-- Device with versions >= Oreo need location permission to start/stop the hotspot-->
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION"/>
     <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION"/>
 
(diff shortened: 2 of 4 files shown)


share_app/src/main/res/values/strings.xml
```diff
@@ -5,7 +5,9 @@
     <string name="stop">Stop</string>
     <string name="hotspot_settings_dialog">Enable portable hotspot from the settings.</string>
     <string name="hotspot_name_suffix">-odk-share</string>
+    <string name="hotspot_name_prefix_oreo">AndroidShare</string>
```
lognaturel: This should probably be ODKShare?

lognaturel: Ah, I see this is merged. Probably not a big deal but something to consider changing at some point!

lakshyagupta21: @lognaturel For oreo devices we can't change its hotspot configuration so when we enable/disable the hotspot, it creates a hotspot with some SSID like `AndroidShare-XXXX` where `XXXX` is some 4 digits random number.

lognaturel: I see! That wasn't clear to me from the issues you linked to. You probably then want to make sure both `hotspot_name_prefix_oreo` and `hotspot_name_suffix` are in an untranslated file rather than in `strings.xml`. It might also be helpful to add a comment to make it clear that this is a value set by Android.

lakshyagupta21: Sure, will make the changes in the PR. Thanks for pointing this out.

PR #60, opened 2018-07-23
Wifi connection action buttons added to main view

### LINKED ISSUE #59: QR code scanner is not easily accessible
On the sending side, the QR code is nice and big and looks like the primary way to initiate the connection. On the receiver side, the scanner is hidden in the overflow menu.

I see a few options:
- use icons for wifi scanning vs QR code scanning so they both fit
- move the two options into the main view so they're more visible
- show the QR code scanner by default with an alternate option to look for networks

lakshyagupta21:
Closes #59 

Both wifi scan and scan QR buttons added to mainview so that they both become more visible. They are not added to menu because QR code scan icon is not used usually,  so some extra info was needed to show the behaviour of that icon.


share_app/src/main/java/org/odk/share/activities/WifiActivity.java
```diff
@@ -395,36 +400,26 @@ private void startReceiveTask() {
...
-
-                integrator.setCaptureActivity(CaptureActivity.class);
-                integrator.setBeepEnabled(true);
-                integrator.initiateScan();
-                return true;
-        }
-        return super.onOptionsItemSelected(item);
+    @OnClick(R.id.bScanQRCode)
+    public void scanQRCode() {
+        IntentIntegrator integrator = new IntentIntegrator(this);
```
shobhitagarwal1612: It would look much cleaner with a chain of calls, wouldn't it?

```
        new IntentIntegrator(this)
                .setPrompt(getString(R.string.scan_qr_code))
                .setDesiredBarcodeFormats(QR_CODE_TYPES)
                .setCameraId(0)
                .setOrientationLocked(false)
                .setCaptureActivity(CaptureActivity.class)
                .setBeepEnabled(true)
                .initiateScan();
```

share_app/src/main/res/layout/activity_wifi.xml
```diff
@@ -10,7 +10,9 @@
     <RelativeLayout
         android:layout_width="match_parent"
         android:layout_height="match_parent"
-        android:layout_below="@id/toolbar">
+        android:layout_below="@id/toolbar"
```
shobhitagarwal1612: Please run `auto-format` or Ctrl + Shift + F (for eclipse key binding) to this file to auto-arrange the lines

lognaturel:
👍 That's great!

PR #61, opened 2018-07-23
Socket connection error's handled

### LINKED ISSUE #58: "Successfully received 0 form(s)" message doesn't make sense
I'm not sure how to reproduce this but it's currently happening consistently with two of my devices. After I pick a filled form to send, the sender is showing the QR code and port number. The receiver successfully connects. The sender doesn't show a dialog but the receiver shows "Successfully received 0 form(s)".

It seems like something went wrong though I don't know what. In general, it seems that if the receiver gets 0 forms, it should show an error message rather than a success message.

lakshyagupta21:
Closes #58 

I was not able to notice any dialog with such behaviour.  But by looking at the code,  it looks like that for any exception or error it returns current progress value of form transfer,  now the exception is caught and sent the ERROR STATUS in UploadEvent or DownloadEvent with error message.

### DIFF
--- a/share_app/src/main/java/org/odk/share/tasks/DownloadJob.java
+++ b/share_app/src/main/java/org/odk/share/tasks/DownloadJob.java
@@ -69,9 +69,7 @@ public class DownloadJob extends Job {
         ((Share) getContext().getApplicationContext()).getAppComponent().inject(this);
 
         initJob(params);
-
-        String result = receiveForms();
-        rxEventBus.post(new DownloadEvent(DownloadEvent.Status.FINISHED, result));
+        rxEventBus.post(receiveForms());
 
         return null;
     }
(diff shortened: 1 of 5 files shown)


share_app/src/main/java/org/odk/share/tasks/DownloadJob.java
```diff
@@ -106,9 +104,13 @@ private String receiveForms() {
 
         } catch (IOException e) {
             Timber.e(e);
+            return new DownloadEvent(DownloadEvent.Status.ERROR, e.getMessage());
+        } catch (IllegalArgumentException e) {
```
shobhitagarwal1612: These catch clauses can be merged

share_app/src/main/java/org/odk/share/tasks/UploadJob.java
```diff
@@ -96,11 +95,13 @@ private String uploadInstances() {
             dis.close();
         } catch (SocketException e) {
             Timber.e(e);
+            return new UploadEvent(UploadEvent.Status.ERROR, e.getMessage());
         } catch (IOException e) {
             Timber.e(e);
+            return new UploadEvent(UploadEvent.Status.ERROR, e.getMessage());
```
shobhitagarwal1612: These catch clauses can be merged

PR #64, opened 2018-07-24
Content added in statistics screen

### LINKED ISSUE #63: Add content to statistics screen
Add various form statistics related to form transfer and review process.

lakshyagupta21:
Closes #63 

Added status of a number of sent, received or reviewed instances.

stats

### DIFF
--- a/share_app/src/main/res/values/strings.xml
+++ b/share_app/src/main/res/values/strings.xml
@@ -68,7 +68,6 @@
     <string name="no_forms_sent">No filled %s forms sent</string>
     <string name="no_forms_received">No filled %s forms received</string>
     <string name="no_forms_reviewed">No finalized %s forms received</string>
-
     <string name="sort_list">Sort the list</string>
     <string name="search_list">Search the list</string>
     <string name="search">Search</string>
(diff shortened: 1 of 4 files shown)


share_app/src/main/res/values/strings.xml
```diff
@@ -68,4 +68,8 @@
     <string name="no_forms_sent">No filled %s forms sent</string>
     <string name="no_forms_received">No filled %s forms received</string>
     <string name="no_forms_reviewed">No finalized %s forms received</string>
+
+    <string name="num_forms_sent">Number of forms sent</string>
```
lognaturel: Maybe consider using "filled forms" here? The context makes it pretty clear but I always think of "form" as "blank form".

shobhitagarwal1612:
I came across this wonderful library which enables us to add rich graphs and charts very easily.
https://github.com/PhilJay/MPAndroidChart

I thought this might add a bit more value to this screen by adding a bit graphics to it.
@lakshyagupta21 @lognaturel What do you say?

lognaturel:
That's really great! Might be useful for some things in Collect as well.

How about filing it as an issue because it is beyond the immediate scope of what we discussed for GSoC success. If there's time, great. But if it has to be added on later, that would be fine as well.

shobhitagarwal1612:
@lakshyagupta21 Can you please open an issue to add the above library in future and resolve the conflicts. Then the PR can be merged

Issue #66, opened 2018-07-26
SQLiteConnection leaked


lognaturel:
On Android 4.1.1 and master, I get a constant stream of the following error in logcat:

`W/SQLiteConnectionPool(12326): A SQLiteConnection object for database '/storage/sdcard0/odk/metadata/forms.db' was leaked!  Please fix your application to end transactions in progress properly and to close the database when it is no longer needed.`

lakshyagupta21:
Fixed in #67. So closing this issue.

PR #67, opened 2018-07-28
Cursor closed

lakshyagupta21:
Cursor was not closed in MainActivity maybe that resulted in `SQLiteConnection Leaked` error.

@lognaturel Please check is it still showing Connection leaked error #66


shobhitagarwal1612:
This can also be replaced with try-with-resources to auto close the cursor after the block gets executed

PR #68, opened 2018-07-28
Showing QR Code only if hotspot is enabled

### LINKED ISSUE #51: If portable hotspot configuration fails, no feedback is given
On two of the devices I've tried (Galaxy Tab A and Alcatel One Touch 5020D, details at https://github.com/opendatakit/collect#contributing-testing), when I tap "Send" and select filled forms, I go straight to a screen that shows "Waiting for connection to accept". What does this state means? It looks like it's not possible to set up a hostpot on those devices? I don't get promtped to enable from the settings.

On the Moto G4 I do get prompted to enable the hotspot from settings. Then I see the QR code and port number information.

lakshyagupta21:
Closes #51 

If a hotspot is enabled then show the QR Code else change the status.

### DIFF
--- a/share_app/src/main/java/org/odk/share/activities/SendActivity.java
+++ b/share_app/src/main/java/org/odk/share/activities/SendActivity.java
@@ -118,8 +118,11 @@ public class SendActivity extends InjectableActivity {
                     switch (hotspotEvent.getStatus()) {
                         case ENABLED:
                             isHotspotRunning = true;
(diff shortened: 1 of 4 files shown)


shobhitagarwal1612:
@lognaturel Can you please verify if this prevents QRCode from getting generated on Android 4.1 device?

lognaturel:
I'm traveling and unfortunately forgot to bring the Android 4.1 device. 😫The code looks good so I'd suggest merging and filing an issue to check at some point.

PR #90, opened 2018-08-19
Disable wifi before triggering the Wifi-Hotspot

### LINKED ISSUE #89: Unable to start hotpot on Android 4.4
Trying to trigger hotspot on Android 4.4 version, it's giving true as a return value when wifi hotspot is triggered.

lakshyagupta21:
Fixes #89 

There is an issue while triggering the wifi-hotspot when a device is already connected to some wifi-network.


lognaturel:
I wasn't able to reproduce the issue. I still can't send from Android 4.1.1 but it doesn't seem to be related to this. The logic makes sense and it doesn't seem to cause any regressions I can find so I have merged it.

Issue #100, opened 2019-02-12
App Crashes On Clicking Send When There are no forms inside


iadeelzafar:
### Summary:
App Crashes On Clicking Send When There are no forms inside

### To Reproduce:
1- Install collect
2- Don't import any form, just go to Share and click on SEND.

I'd like to work on this issue. Suggestions are welcomed on how this issue could be resolved.

### Screenshot:
<img src="https://user-images.githubusercontent.com/36811908/52628624-dfdb8f80-2ed9-11e9-949a-8c783287888d.gif" height="700" width="394">

shobhitagarwal1612:
I guess SEND button shouldn't be enabled at all if no forms are available
@lakshyagupta21 any thoughts?

lakshyagupta21:
@shobhitagarwal1612 correct it shouldn't be enabled.
Expected behavior - `send` button should be disabled as well as there should be some empty view message on the screen.

iadeelzafar:
@lakshyagupta21 I opened an issue before about the empty view too.

iadeelzafar:
@opendatakit-bot claim

PR #106, opened 2019-02-19
Add Ability to Configure ODK Destination Dir

### LINKED ISSUE #105: Configurable ODK submissions destination folder
The ODK submissions destination folder is not configurable, can we make it configurable?

 A user may not always want the submission instances in the same folder, for example, to differentiate submissions from different devices when synchronised. Having the user set the destination folder would allow them to be able to delete data from other devices. This stems from our use case, where:
1. A group of enumerators in the field collect data
2. After every X enumerations, they meet, synchronise their submissions using ODK share
3. A separate tool checks coverage based on indicators in all submissions including synchronised submissions of the enumerators in the field (Note: data may not have been submitted to an ODK aggregate server yet, they could still be in the field with no internet access.)
4. At the end of the day, each enumerator only submits the data they collected to ODK Aggregate server.

1 - 3 only happens while enumerators are in the field, and they may be offline still and may not have submitted their data to an aggregate server.

rkodev:
Add the ability to change which directory files from other devices are
downloaded to.

Fixes  #105 

Signed-off-by: Jason Rogena
Signed-off-by: Ronald Kudoyi


shobhitagarwal1612:
@rkodev Did you face the above mentioned problem on your device?

rkodev:
@shobhitagarwal1612 yes, but I believe this issue has already been reported and worked in here #100

shobhitagarwal1612:
> If I send a blank form and use a different destination folder, then in ODK Collect the form gets loaded properly but the media isn't as it still tries to look for the media in sdcard/odk/ folder

This bug is in ODK Collect side and is still present. Did you try to open the transferred instance after pushing to a different directory?

shobhitagarwal1612:
Please resolve the conflicts. This PR can be merged then

shobhitagarwal1612:
I have resolved the conflicts here https://github.com/opendatakit/skunkworks-crow/pull/190

Closing this

shobhitagarwal1612:
The PR has been merged. Thanks for helping to improve the project!!!

PR #107, opened 2019-02-19
4 prevent duplicating of submissions in destination

rkodev:
Use Instance UUID as the True Source Of Truth

Switch from using the instance ID to the instance UUID (that's now a SHA1 hash of the device ID + the instance directory, instead of a random number) as the true reference to a transfer, instead of using the instance ID.

### DIFF
--- a/share_app/src/main/java/org/odk/share/application/Share.java
+++ b/share_app/src/main/java/org/odk/share/application/Share.java
@@ -28,7 +28,7 @@ public class Share extends DaggerApplication {
         return singleton;
     }
 
-    public static final String ODK_ROOT = Environment.getExternalStorageDirectory() + File.separator + "share";
+    public static final String ODK_ROOT = Environment.getExternalStorageDirectory() + File.separator + "odk";
     public static final String ODK_COLLECT_ROOT = Environment.getExternalStorageDirectory() + File.separator + "odk";
     public static final String FORMS_PATH = ODK_COLLECT_ROOT + File.separator + "forms";
     public static final String INSTANCES_PATH = ODK_COLLECT_ROOT + File.separator + "instances";
--- a/share_app/src/main/java/org/odk/share/utilities/OdkInstanceUtils.java
+++ b/share_app/src/main/java/org/odk/share/utilities/OdkInstanceUtils.java
@@ -1,0 +1,48 @@
+package org.odk.share.utilities;
+
+import org.odk.share.application.Share;
+
+import java.io.UnsupportedEncodingException;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+
+public class OdkInstanceUtils {
+    /**
+     * Uniquely identifiable ID for an instance would be a combination of the name of the directory
+     * for the instance and the device id.
+     * @return
+     */
+    public static String getInstanceUuid(String instancePath) throws UnsupportedEncodingException, NoSuchAlgorithmException {
+        return SHA1(getDeviceId() + instancePath);
+    }
+
+    private static String convertToHex(byte[] data) {
+        StringBuilder buf = new StringBuilder();
+        for (byte b : data) {
+            int halfbyte = (b >>> 4) & 0x0F;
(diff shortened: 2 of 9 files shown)


share_app/src/main/java/org/odk/share/application/Share.java
```diff
@@ -28,7 +28,7 @@ public static Share getInstance() {
         return singleton;
     }
 
-    public static final String ODK_ROOT = Environment.getExternalStorageDirectory() + File.separator + "share";
+    public static final String ODK_ROOT = Environment.getExternalStorageDirectory() + File.separator + "odk";
```
shobhitagarwal1612: Changing this would move the share's database `share.db` from `/sdcard/share/` to `/sdcard/odk/`

Is this change really required for fixing this issue?

shobhitagarwal1612: @rkodev Do you agree with the above statement?

rkodev: @shobhitagarwal1612  yes, we can have this reverted back to the share directory

shobhitagarwal1612:
@lakshyagupta21 Please test it and merge if this looks good to you

share_app/src/main/java/org/odk/share/utilities/OdkInstanceUtils.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    /**
+     * Same logic used within ODK Collect to obtain the device id in a form
+     *
+     * @return String
+     */
+    private static String getDeviceId() {
+        PropertyManager propertyManager = new PropertyManager(Share.getInstance());
+        return propertyManager.getSingularProperty(PropertyManager.DEVICE_ID_PROPERTY);
```
lakshyagupta21: Is it okay if we use device-specific information without user consent?

rkodev: @lakshyagupta21 this id never leaves the device as its encrypted. The only access to this method is via a getter `org.odk.share.utilities.OdkInstanceUtils#getInstanceUuid` which is a SHA1 encrypted algorithm

shobhitagarwal1612:
The conflicts needs to be resolved

shobhitagarwal1612:
@lakshyagupta21 We need to decide whether to accept this change or not. If I understood correctly, your argument is that user-consent shouldn't be involved in this. Am I correct? If not, then let's discuss it and get to a conclusion

lakshyagupta21:
> . If I understood correctly, your argument is that user-consent shouldn't be involved in this. Am I correct?

Correct.

shobhitagarwal1612:
Ok, so what is the advantage of going with this approach as compared to the current implementation except for the user-consent point?

share_app/src/main/java/org/odk/share/utilities/OdkInstanceUtils.java
jasonrogena: @lakshyagupta21 the reason we used a hash of the device ID was to guarantee that there wouldn't be collisions in the directory names for the submissions (in the receiving device) and at the same time we're able to always recreate the same directory name for a submission. We could, however, switch up the logic; maybe generate UUIDs for the submissions. We'll have to save these in share's SQLite database though.

ukanga: Could we use the UUID already in the submission.xml? Or will that introduce more overhead by having to read the submission XML to get the UUID for each submission?

rkodev: @ukanga The new UUID we are proposing will be the external 'Device ID'  instead of using the actual device id. Its to be used by the receiving device to track all submissions by the same device to control duplication.
The submission UUID is generated with every new submission so its impossible to use it to track the submissions by the same device. That's why we want to store a new 'Device ID' at the same time addressing @lakshyagupta21 concerns

PR #110, opened 2019-02-20
Solves NullPointerException and Empty view problem.

### LINKED ISSUE #109: Problems after first time installation.
#### Problem Description
There are some problems at initial setup of the app. With initial setup, I mean there is no data in  collect as well as in share. 
1. It will not show the Empty view TextView with text "No Blank forms available". 
2. On clicking send button when there is no forms available, will result in crash. As reported in #100.

These two problems are arising, because at first time use, the cursor will be null.

**Cause of 1st problem** : the method ```setEmptyViewVisibility()```is skipped all the time as it is only called when cursor is not equal to null. (Same behaviour at FilledFormsFragment and BlankFormsFragment)

screenshot_2019-02-20-23-48-23-79 2

**Cause of 2nd problem** : in onLoadFinished() method in FilledFormsFragment and BlankFormsFragment, cursor is directly accessed without null check.

> java.lang.NullPointerException: Attempt to invoke interface method 'boolean android.database.Cursor.moveToFirst()' on a null object reference
        at org.odk.share.fragments.FilledFormsFragment.onLoadFinished(FilledFormsFragment.java:93)
        at org.odk.share.fragments.FilledFormsFragment.onLoadFinished(FilledFormsFragment.java:39)

#### Steps To Reproduce:

1. Delete all the data of collect and share.
2. Reinstall both the apps in your device.
3. Don't import any blank forms in collect.
4. Launch the share app, you will find all these problems.

#### Other Informations
Disabling the send button as proposed in #102, when there are no forms, can solve 2nd problem but it will not ensure the data consistency as cursor is accessed without null check.

DreamyPhobic:
Closes #109 

#### Changes introduced:
1. Added Null Check for the cursor.
2. Called ```setEmptyViewVisibility()``` with 0 as argument in case, when cursor is null.

If there is some better solution, please let me know.

### DIFF
--- a/share_app/src/main/java/org/odk/share/activities/MainActivity.java
+++ b/share_app/src/main/java/org/odk/share/activities/MainActivity.java
@@ -138,11 +138,11 @@ public class MainActivity extends FormListActivity implements LoaderManager.Load
     @Override
     public void onLoadFinished(@NonNull Loader<Cursor> loader, Cursor cursor) {
         formAdapter.changeCursor(cursor);
-        if (cursor != null) {
-            if (!cursor.isClosed())
-                setEmptyViewVisibility(cursor.getCount());
-        } else
-            setEmptyViewVisibility(0);
+        if (cursor != null && !cursor.isClosed()) {
+            setEmptyViewVisibility(cursor.getCount());
+            return;
+        }
+        setEmptyViewVisibility(0);
     }
 
     @Override
--- a/share_app/src/main/java/org/odk/share/activities/MainActivity.java
+++ b/share_app/src/main/java/org/odk/share/activities/MainActivity.java
@@ -138,9 +138,11 @@ public class MainActivity extends FormListActivity implements LoaderManager.Load
     @Override
     public void onLoadFinished(@NonNull Loader<Cursor> loader, Cursor cursor) {
         formAdapter.changeCursor(cursor);
-        if (cursor != null && !cursor.isClosed()) {
-            setEmptyViewVisibility(cursor.getCount());
-        }
+        if (cursor != null) {
+            if (!cursor.isClosed())
+                setEmptyViewVisibility(cursor.getCount());
+        } else
+            setEmptyViewVisibility(0);
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


share_app/src/main/java/org/odk/share/activities/MainActivity.java
```diff
@@ -138,9 +138,11 @@ protected String getSortingOrderKey() {
     @Override
     public void onLoadFinished(@NonNull Loader<Cursor> loader, Cursor cursor) {
         formAdapter.changeCursor(cursor);
-        if (cursor != null && !cursor.isClosed()) {
-            setEmptyViewVisibility(cursor.getCount());
-        }
+        if (cursor != null) {
```
lakshyagupta21: It could be like
```
if (cursor != null && !cursor.isClosed()) {
    setEmptyViewVisibility(cursor.getCount());
    return;
}
setEmptyViewVisibility(0);
```

share_app/src/main/java/org/odk/share/fragments/FilledFormsFragment.java
```diff
@@ -90,16 +90,19 @@ public void onResume() {
...
+            recyclerView.setAdapter(instanceAdapter);
+            setEmptyViewVisibility(cursor.getCount());
+            if (instanceAdapter.getItemCount() > 0) {
+                toggleButton.setText(getString(R.string.select_all));
+                toggleButton.setEnabled(true);
+            } else {
+                toggleButton.setEnabled(false);
+            }
+        } else
+            setEmptyViewVisibility(0);
```
lakshyagupta21: Make sure to write every block in braces `{}` even if the block contains a single line statement.

lakshyagupta21:
Thanks @DreamyPhobic for working on the PR.
One suggestion that I would like to give on your commit messages, always try to write commit messages related to the changes that you have done in that commit, so that other developers get some idea about the changes, like your last commit message is similar to the first one but you fixed lint checkstyle or pmd issues in second commit, so your commit message should be like `fixed lint/checkstyle/pmd issues` or anything else which describes briefly about the change.

DreamyPhobic:
thanks, I will take care from next time.

PR #116, opened 2019-02-23
Inject Dao's in activities/fragments using Dagger

### LINKED ISSUE #115: Remove singleton pattern from Application class
`Share` application class stores a static reference to the application's context. We should get rid of it and use the context available to us via fragment/activity

shobhitagarwal1612:
Closes #115 

 - Removes static application's instance and it's usage from everywhere
 - Makes activities/fragments injectable
 - Let dagger handle injection for Dao instead of creating new instances manually


shobhitagarwal1612:
@lakshyagupta21 please review

share_app/src/main/java/org/odk/share/adapters/FormsAdapter.java
```diff
@@ -29,16 +29,20 @@
 
 public class FormsAdapter extends CursorRecyclerViewAdapter<FormsAdapter.FormHolder> {
 
+    private final InstancesDao instancesDao;
+    private final TransferDao transferDao;
     @Nullable
-    private LinkedHashSet<Long> selectedForms;
+    private final LinkedHashSet<Long> selectedForms;
```
lakshyagupta21: Any reason for making it final?

lakshyagupta21:
And there are some merge conflicts as well.

share_app/src/main/java/org/odk/share/adapters/FormsAdapter.java
shobhitagarwal1612: To ensure that the variable is never mutated again in the class.

shobhitagarwal1612:
Fixed

PR #161, opened 2019-03-02
Fixes #157 Empty Dialog if blank form already available

### LINKED ISSUE #157: Empty transfer dialog if blank form already available

#### Problem description
When transferring blank forms if all are already available at the receiver, then the resulting dialog is blank

Send blank forms which are already available in the receiver

#### Expected behavior
Result dialog should display that `Forms are already available` or something similar to indicate what has happened.

Screenshot:
image

rohap97:
Closes #157

### DIFF
--- a/share_app/src/main/java/org/odk/share/activities/WifiActivity.java
+++ b/share_app/src/main/java/org/odk/share/activities/WifiActivity.java
@@ -168,8 +168,8 @@ public class WifiActivity extends InjectableActivity implements OnItemClickListe
                             break;
                         case FINISHED:
                             dismissDialog(DIALOG_DOWNLOAD_PROGRESS);
-                            //String result = downloadEvent.getResult();
-                            createAlertDialog(getString(R.string.transfer_result), getString(R.string.finished_downloading));
+                            String result = downloadEvent.getResult();
+                            createAlertDialog(getString(R.string.transfer_result), result);
                             break;
                         case ERROR:
                             dismissDialog(DIALOG_DOWNLOAD_PROGRESS);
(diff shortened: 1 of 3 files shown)


lakshyagupta21:
@rohap97 Please follow the PR template this helps in the review.

rohap97:
@shobhitagarwal1612  Sir, help needed regarding this issue.I am unable to get an instance of the downloaded form and compare it with the existing form , How can I do this? In DownloadJob.java there is a method called isFormExits(), it might be able to tell if the form exists or not, but I am not certain. Please help

lakshyagupta21:
@rohap97 UploadJob sends blank form information to DownloadJob at the receiver end which checks whether any form with the received information exists at receiver end or not. If it exists then it sends some flag to sender's UploadJob that form already exists with me, send another if you still have something to send and if  form doesn't then also it sends some flag that the form with given information doesn't exist at my end so send it. This is how flow works.
So you have to change in the UploadJob file where it checks for flag `formExistAtReceiver`

rohap97:
@lakshyagupta21 Sir, I have understood the flow of work.  I am still getting lost on how to be able to use the flag formExistAtReciever in WifiActivity.java. Please help

lakshyagupta21:
@rohap97 You've to add string result of every form to `sbResult` variable for every result. This will be displayed in the dialog directly.
You can use `sbResult.append(formName + " "  + formVersion + " already available at the receiver) ` whenever form exists at the receiver and for which check is already there.
But whatever string you append to `sbResult `always create string resource in `string.xml` its easier to reuse and change directly from there.

rohap97:
This PR was closed because i made a few mistakes and was unable to rectify them.

lakshyagupta21:
@rohap97 If you face any difficulty using Git or with anything else just ask questions either here or on slack but don't close PRs.

rohap97:
@lakshyagupta21 Ok Sir, will not do it again.

PR #170, opened 2019-03-03
Contributing guidelines added and readme updated

lakshyagupta21:
Contributing.md template is used from ODK-Collect and made some modifications according to the project and readme is also updated with the link.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,8 +31,6 @@ To contribute code to Skunkworks-Crow, you will need to open a [pull request](ht
 1. Push changes to your fork at any time to make them publicly available:
 
         git push
-        
-1. Once you have completed your code changes, verify that you have followed the [style guidelines](https://github.com/opendatakit/skunkworks-crow/blob/master/CONTRIBUTING.md#style-guidelines).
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,95 @@
...
+
+Small fixes that target very particular bugs may occasionally be merged without a second review.
+
+## Work in progress pull requests
+
+Work in progress (WIP) pull requests are useful to illustrate a proposed direction and get early feedback before committing to a solution direction.
```
shobhitagarwal1612: Now that GitHub itself has a feature to make a pull request a work in progress, I think we should add that bit as well

CONTRIBUTING.md
```diff
@@ -0,0 +1,95 @@
...
+
+1. Once you've made incremental progress towards you goal, commit your changes with a meaningful commit message. Use [keywords for closing issues](https://help.github.com/articles/closing-issues-via-commit-messages/) to refer to issues and have them automatically close when your changes are merged.
+
+        git commit -m "Do a thing. Fix #1."
+
+1. Push changes to your fork at any time to make them publicly available:
+
+        git push
+        
+1. Once you have completed your code changes, verify that you have followed the [style guidelines](https://github.com/opendatakit/skunkworks-crow/blob/master/CONTRIBUTING.md#style-guidelines).
```
shobhitagarwal1612: I don't think we have a style guideline in this repo. So should be removed

PR #171, opened 2019-03-03
Add Up Button In Activities For Better Navigation

### LINKED ISSUE #108: Add Up Button In SendFormsActivity and WifiActivity
### Describe the feature:

* Adding an up button in SendFormsActivity and WifiActivity.
* What you guys think? @shobhitagarwal1612 @lakshyagupta21

iadeelzafar:
Closes #108 

#### Changes:
- [x] Added up button in WifiActivity
- [x] Added up button in AboutActivity
- [x] Added up button in InstanceManagerTabs
- [x] Added up button in Send Forms Activity

Tested it in circumstances such as:
* Clicking on the up button.
* Pressing back

This is the best possible solution according to [Android Documentation](https://developer.android.com/training/appbar/up-action)
No risks are associated with these changes. 

### DIFF
--- a/skunkworks_crow/src/main/AndroidManifest.xml
+++ b/skunkworks_crow/src/main/AndroidManifest.xml
@@ -33,11 +33,23 @@
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
         </activity>
-        <activity android:name=".activities.WifiActivity" />
+
+        <activity android:name=".activities.WifiActivity"
+                  android:parentActivityName=".activities.MainActivity">
+            <meta-data
+                    android:name="android.support.PARENT_ACTIVITY"
+                    android:value=".activities.MainActivity"/>
+        </activity>
         <activity android:name=".activities.InstancesList" />
         <activity android:name=".preferences.SettingsPreference" />
         <activity android:name=".activities.WebViewActivity" />
-        <activity android:name=".activities.AboutActivity" />
+
+        <activity android:name=".activities.AboutActivity"
+                  android:parentActivityName=".activities.MainActivity">
+            <meta-data
+                    android:name="android.support.PARENT_ACTIVITY"
+                    android:value=".activities.MainActivity"/>
+        </activity>
         <activity
             android:name=".activities.SendActivity"
             android:configChanges="orientation|screenSize" />
@@ -49,7 +61,14 @@
             android:screenOrientation="portrait"
             android:stateNotNeeded="true"
             tools:replace="android:screenOrientation" />
-        <activity android:name=".activities.InstanceManagerTabs" />
+
+        <activity android:name=".activities.InstanceManagerTabs"
+                  android:parentActivityName=".activities.MainActivity">
+            <meta-data
+                    android:name="android.support.PARENT_ACTIVITY"
+                    android:value=".activities.MainActivity"/>
+        </activity>
+
         <activity android:name=".activities.ReviewFormActivity" />
 
         <provider
@@ -59,7 +78,13 @@
             android:name=".provider.InstanceMapProvider"
             android:authorities="org.odk.share.provider.odk.map" />
 
-        <activity android:name=".activities.SendFormsActivity"></activity>
+        <activity android:name=".activities.SendFormsActivity"
+                  android:parentActivityName=".activities.MainActivity">
+            <meta-data
+                    android:name="android.support.PARENT_ACTIVITY"
+                    android:value=".activities.MainActivity"/>
+        </activity>
+
     </application>
 
 </manifest>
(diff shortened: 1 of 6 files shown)


iadeelzafar:
Up button could not be added in the SettingsPreference as it extends from PreferenceActivity.

``` public class SettingsPreference extends PreferenceActivity { ```

We can't extend it from AppCompatActivity too as multiple inheritance is not allowed in Java. I searched on StackOverFlow but couldn't find a solution for such a case.

**Comment by iadeelzafar** [comment]
@Chromicle Thanks for reviewing.
ab is the variable name for ActionBar which seems accurate as it follows the Android's official documentation example's naming convention. Similarly, you can see variable names such as llm for LinearLayoutManager.
Secondly, read my above comment that is related to SettingsPreference.

File: skunkworks_crow/src/main/java/org/odk/share/activities/AboutActivity.java [hunk 1]
```diff
@@ -37,6 +38,10 @@ protected void onCreate(Bundle savedInstanceState) {
         setTitle(getString(R.string.about));
         setSupportActionBar(toolbar);
 
+        setSupportActionBar(toolbar);
+        ActionBar ab = getSupportActionBar();
```
lakshyagupta21: can directly be written as `getSupportActionBar().setDisplayHomeAsUpEnabled(true);`

skunkworks_crow/src/main/java/org/odk/share/activities/AboutActivity.java
```diff
@@ -37,6 +38,10 @@ protected void onCreate(Bundle savedInstanceState) {
         setTitle(getString(R.string.about));
         setSupportActionBar(toolbar);
 
+        setSupportActionBar(toolbar);
```
lakshyagupta21: This is not needed, please have a look at line no 39

skunkworks_crow/src/main/AndroidManifest.xml
```diff
@@ -37,7 +37,12 @@
         <activity android:name=".activities.InstancesList" />
         <activity android:name=".preferences.SettingsPreference" />
         <activity android:name=".activities.WebViewActivity" />
-        <activity android:name=".activities.AboutActivity" />
+        <activity android:name=".activities.AboutActivity"
```
lakshyagupta21: Add `android:name=".activities.AboutActivity"` in seperate line.

iadeelzafar:
@lakshyagupta21 Thanks for reviewing and suggesting the changes. I've made these changes. Have a look.

shobhitagarwal1612:
The changes are definitely an improvement. I believe that the issue was to fix `SendFormsActivity and WifiActivity`, so can you fix that as well since merging this PR would close issue #108 

Thanks for the great work and helping in reviewing other PRs

iadeelzafar:
@shobhitagarwal1612 Thanks for writing. I've made changes to the WifiActivity just now.
I tried adding upButton to the SendFormsActivity but was unable to do so because it uses the actionBar from the MainActivity which is not being handled in the ``` SendFormsActivity ``` itself. I looked up on StackOverFlow for finding a solution for such a case but couldn't find one.

shobhitagarwal1612:
Can you add a checklist to the PR description containing a list of activities that need to be fixed?

iadeelzafar:
@shobhitagarwal1612 I've added a checklist to show the activities in which I've added the up button.

shobhitagarwal1612:
> Send Forms Activity

What issues were you facing when adding a back button for this activity aka `InstanceManagerTabs` ?

iadeelzafar:
@shobhitagarwal1612 Yeah. I was trying a different approach but this one has worked. I agree that there should be a uniform approach in all activities. I've made changes for the SendFormsActivity.

iadeelzafar:
@shobhitagarwal1612 The only issue now is with the SettingsPreference as it inherits from ```PreferenceActivity``` so, methods like ```setSupportActionBar()``` and ```getSupportActionBar()``` don't work here. I'm trying to find an alternative if there is any.

shobhitagarwal1612:
In Preference activity, the toolbar is being inflated and then added to the layout. See line 34 in `SettingsPreference`

iadeelzafar:
@shobhitagarwal1612 I noticed that earlier and tried working with the toolbar but since SettingsPreference extends from PreferenceActivity so we can't use methods such as set and getSupportActionBar().
See the second answer here, [Stack Overflow](https://stackoverflow.com/questions/37222879/add-action-bar-with-back-button-in-preference-activity)

iadeelzafar:
@shobhitagarwal1612 Please have a look at my above comment.

iadeelzafar:
@shobhitagarwal1612 I tried adding up button in the ```SettingsPreference``` using the provided [solutions](https://stackoverflow.com/questions/26564400/creating-a-preference-screen-with-support-v21-toolbar) and tried on my own too but nothing seems to be promising. Regression issues were coming up. 
So far, I've added an up button to 4 activities. PR description has the details. I've updated it.
I think we can move forward with this.

PR #183, opened 2019-03-06
Updated UI the bottom bar layout, so it looks good

### LINKED ISSUE #174: Update UI of bottam bar instead of direct buttons
#### Problem description
Currently, for a bottom bar, there are directly buttons are there make a better UI and propose below the comments section
the better UI holds for a better user experience so there should be good UI 

#### Expected behavior
good UI design so it holds for good user experience

Chromicle:
Closes #174 

## What has been done to verify that this works as intended?
Tested in Android Studio and my phone(Samsung j8), it looks good as I expected 

####  GIF
UI

#### Why is this the best possible solution
Made 2 drawable resource files so 2 buttons look different and applied the background to the buttons also. and added the same layout to fill forms and blank forms and main activity also

As by updating the UI so it holds good for user experience and so the app looks like beautiful   


iadeelzafar:
* Also, the background color is not the primary color, using the exact same darker color would make it look even nicer.

Chromicle:
@shobhitagarwal1612 I updated as per suggestions can you please review now

Chromicle:
I had a thought that implemented FAB that scrolls to top as if the number of forms is large if the user wants to come up he has to scroll to the top. soi can I raise the issue for that

shobhitagarwal1612:
The buttons need to be consistent everywhere in the app. Right now, they look different on some screens and the padding/margin is different on some. Also, the selected state of the button doesn't look very neat. @lakshyagupta21 What are your thoughts on this?

lakshyagupta21:
I agree with @shobhitagarwal1612 , that whatever behavior we intend to choose it should be consistent on all the screens. And the selected state (color when a button is tapped) should also be similar to the app's primary or secondary color this makes UI consistent everywhere. And by looking at the GIF it looks like there is a lot of padding added in button layout which doesn't look good.

Chromicle:
> I agree with @shobhitagarwal1612 , that whatever behavior we intend to choose it should be consistent on all the screens. And the selected state (color when a button is tapped) should also be similar to the app's primary or secondary color this makes UI consistent everywhere. And by looking at the GIF it looks like there is a lot of padding added in button layout which doesn't look good.

Yeah I agree with @shobhitagarwal1612 I will update it to all screens and I will decrease the padding also

Chromicle:
@shobhitagarwal1612 @lakshyagupta21 now I updated  to all buttons to look familiar

lakshyagupta21:
@Chromicle If you would take a look at the above screenshot, there is a lot of space between the buttons as well as lot of space between buttons and left and right margins.





@Chromicle And in above two screenshots, you can see the inconsistent behaviour in both the screens. And one more thing is that the button layout(layout which contains both buttons) is white which I think looks quite odd. We can make the button layout transparent as well as make button background to white.

lakshyagupta21:
@Chromicle Also `send` button background is grey by default, please take a look. And I would suggest to change that grey color to something similar to primary or secondary color of app.

Chromicle:
Yeah I will make the changes while Iam making the layout weight =0 build is failing 
and I had some exams over a week so I will update after 1 week if I got time I will update it as soon as possible
Please did not assign to any others

lakshyagupta21:
No hurry,  you focus on your exams first,  that is more important.

PR #188, opened 2019-03-09
Buttons removed in "Sent" tab

### LINKED ISSUE #184: no action takes place on tapping the buttons

#### Problem description
When tapped on "Send" and "Select all", no action takes place.

sent

#### Expected behavior
Maybe some activity can be associated with the respective buttons **OR**
 it can be removed from there.
#### Other information 

iamvasuh:
Closes #184 

I made the app run on a device and it's showing the expected result.
The two buttons "Send" and "Select All" are removed as there was no use of them.

This was the best possible solution for the result to be seen and redundancy to be eliminated

It improves the UI/UX.
sent1

- [ ] run `./gradlew pmd checkstyle lint findbugs` and confirmed all checks still pass OR confirm CircleCI build passes

### DIFF
--- a/skunkworks_crow/src/main/java/org/odk/share/fragments/SentInstancesFragment.java
+++ b/skunkworks_crow/src/main/java/org/odk/share/fragments/SentInstancesFragment.java
@@ -162,7 +162,7 @@ public class SentInstancesFragment extends InstanceListFragment {
 
     private void setEmptyViewVisibility(String text) {
         if (transferInstanceList.size() > 0) {
-            buttonLayout.setVisibility(View.VISIBLE);
+            buttonLayout.setVisibility(View.INVISIBLE);
             emptyView.setVisibility(View.GONE);
         } else {
             buttonLayout.setVisibility(View.GONE);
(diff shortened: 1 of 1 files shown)


iamvasuh:
@shobhitagarwal1612  @lakshyagupta21 plz have a look.

skunkworks_crow/src/main/java/org/odk/share/fragments/SentInstancesFragment.java
```diff
@@ -162,7 +162,7 @@ private void setupAdapter() {
 
     private void setEmptyViewVisibility(String text) {
         if (transferInstanceList.size() > 0) {
-            buttonLayout.setVisibility(View.VISIBLE);
+            buttonLayout.setVisibility(View.INVISIBLE);
```
shobhitagarwal1612: If button layout is not going to be visible in this fragment, then why not do it just once in onCreateView?

iamvasuh: there will be 2-3 lines of code to be removed if I do it from onCreate

shobhitagarwal1612: I agree that your choice of editing least number of lines was reasonable. But that creates a redundancy which must be resolved

iamvasuh: yes sure

iamvasuh:
@shobhitagarwal1612 @lakshyagupta21  plz have a look on the update

PR #200, opened 2019-03-10
checkstyle plug-in  is updated

### LINKED ISSUE #192: Update checkstyle plugin
#### Problem description
We haven't upgraded the `checkstyle` plugin for quite some time. Now, we should use the latest version to include all of the new features added recently

Not an issue, code maintenance

#### Expected behavior
`quality.gradle` contains the configuration for `checkstyle` plugin, that should be upgraded. All of the introduced modules should be added to `checkstyle.xml`

#### Other information 

iamvasuh:
Closes #192 

I have updated the **checkstyle** plug-in version present in **_quality.gradle_**  file from **7.6.1** to **8.18** which is the latest one right now.

Apart from that , in the **_checkstyle.xml_** file I have added the modules which are used in **_google_checks.xml_** also.

This is perhaps the best solution according to me.

1.) It helps in keeping the project files up-to-date.
2.) There is proper code maintenance.

### DIFF
--- a/config/quality.gradle
+++ b/config/quality.gradle
@@ -24,7 +24,7 @@ apply plugin: 'checkstyle'
 checkstyle.toolVersion = '8.18'
 
 task checkstyle(type: Checkstyle) {
-    configFile file("$configDir/checkstyle.xml")
+    configDir file("$configDir/checkstyle.xml")
 
     ignoreFailures false // Fail early
     showViolations true
--- a/config/quality.gradle
+++ b/config/quality.gradle
@@ -21,7 +21,7 @@ def reportsDir = "${project.buildDir}/reports"
 
 apply plugin: 'checkstyle'
 
-checkstyle.toolVersion = '7.6.1'
+checkstyle.toolVersion = '8.18'
 
 task checkstyle(type: Checkstyle) {
     configFile file("$configDir/checkstyle.xml")
--- a/config/quality.gradle
+++ b/config/quality.gradle
@@ -21,6 +21,8 @@ def reportsDir = "${project.buildDir}/reports"
 
 apply plugin: 'checkstyle'
 
+checkstyle.toolVersion = '8.18'
+
 task checkstyle(type: Checkstyle) {
     configFile file("$configDir/checkstyle.xml")
 
@@ -102,4 +104,9 @@ android {
         htmlOutput file("$reportsDir/lint/lint-result.html")
         xmlOutput file("$reportsDir/lint/lint-result.xml")
     }
+}
+
+//------------------------ ALL ------------------------//
+task checkCode(type: GradleBuild) {
+    tasks = ['pmd', 'checkstyle', 'lint', 'findbugs']
 }
--- a/config/quality.gradle
+++ b/config/quality.gradle
@@ -21,7 +21,7 @@ def reportsDir = "${project.buildDir}/reports"
 
 apply plugin: 'checkstyle'
 
-checkstyle.toolVersion = '8.18'
+checkstyle.toolVersion = '7.6.1'
 
 task checkstyle(type: Checkstyle) {
     configFile file("$configDir/checkstyle.xml")
--- a/config/quality.gradle
+++ b/config/quality.gradle
@@ -21,8 +21,6 @@ def reportsDir = "${project.buildDir}/reports"
 
 apply plugin: 'checkstyle'
 
-checkstyle.toolVersion = '7.6.1'
-
 task checkstyle(type: Checkstyle) {
     configFile file("$configDir/checkstyle.xml")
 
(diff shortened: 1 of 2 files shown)


iamvasuh:
@shobhitagarwal1612  @lakshyagupta21  plz have a look

iamvasuh:
@shobhitagarwal1612 @lakshyagupta21 plz have a look

lakshyagupta21:
@iamvasuh The main aim of this issue is to update the checkstyle plugin version, so I think adding other module dependencies in the same is not a good idea. We can add those as and when required.
What do you think @shobhitagarwal1612 ?

huangyz0918:
I'm not sure why you have so many changes @iamvasuh , but mentioned by @shobhitagarwal1612 from issue description:
>quality.gradle contains the configuration for checkstyle plugin, that should be upgraded. All of the introduced modules should be added to checkstyle.xml

From here I think adding is expected behavior, no problem, what do you think? @lakshyagupta21

iamvasuh:
> I'm not sure why you have so many changes @iamvasuh , mentioned by @shobhitagarwal1612 from issue description:
> 
> > quality.gradle contains the configuration for checkstyle plugin, that should be upgraded. All of the introduced modules should be added to checkstyle.xml
> 
> I think we don't need set the specific checkstyle version, just remove the version configuration, it will make the checkstyle version latest.


Yes @huangyz0918 @lakshyagupta21  i was also skeptical about adding new modules by myself but i  noticed that, in the **_checktyle.xml_** file, most of the modules were same as that of **_google_checks.xml_** file  and that was the reason I updated all of them.

I will just change the following.

iamvasuh:
Now i have only updated the **checkstyle** tool version and kept the **_checkstyle.xml_**  file  same as it was before. @lakshyagupta21 @huangyz0918 @shobhitagarwal1612

shobhitagarwal1612:
Something has gone wrong. All of the changes are removed. Please fix

iamvasuh:
> Something has gone wrong. All of the changes are removed. Please fix

yes sure

iamvasuh:
@shobhitagarwal1612 please help me out in understanding what i am doing wrong in updating the checkstyle version. I have already tried the following ways

1. I had updated the version from 7.6.1 to 8.18 , and had changed the checkstyle.xml file content with that of google_checks.xml.

2. after the suggestion of @huangyz0918 , I tried to implement his suggestion but it seems that i am not able to apply it properly.

rohap97:
@iamvasuh Try to make the curly braces of if-else block in the same line. It is because of this error that the build is failing. The associated file is InstancesList.java, error is at Line no. 17. For example it should be like this "} else {".

iamvasuh:
> @iamvasuh Try to make the curly braces of if-else block in the same line. It is because of this error that the build is failing. The associated file is InstancesList.java, error is at Line no. 17. For example it should be like this "} else {".

actually i have a doubt regarding the update of the plugin

huangyz0918:
@iamvasuh removing the checkstyle version would keep the checkstyle version latest according to the supported Gradle version, for example, from Gradle 3.3 onwards, the default checkstyle version is 6.19; before, it was 5.9. Only Gradle versions prior to 2.4 used even older versions of checkstyle. 

So, as we mentioned above, we can use the default checkstyle version, it will auto update as our Gradle version updates, or, we can just update the checkstyle version to `8.18` like your [commit](https://github.com/opendatakit/skunkworks-crow/pull/200/commits/859e8ecc4f33dcc225ccf250891533d4bb906ba2). We can discuss the two options with @lakshyagupta21 and @shobhitagarwal1612 and choose one.

This is not a big issue, if you use the default checkstyle version, I see there are three violations in `activities/InstancesList.java`, `BlankFormsFragment.java` and `FilledFormsFragment.java`. Fix them and get the PR passed.

iamvasuh:
> @iamvasuh removing the checkstyle version would keep the checkstyle version latest according to the supported Gradle version, for example, from Gradle 3.3 onwards, the default checkstyle version is 6.19; before, it was 5.9. Only Gradle versions prior to 2.4 used even older versions of checkstyle.
> 
> So, as we mentioned above, we can use the default checkstyle version, it will auto update as our Gradle version updates, or, we can just update the checkstyle version to `8.18` like your [commit](https://github.com/opendatakit/skunkworks-crow/pull/200/commits/859e8ecc4f33dcc225ccf250891533d4bb906ba2). We can discuss the two options with @lakshyagupta21 and @shobhitagarwal1612 and choose one.
> 
> This is not a big issue, if you use the default checkstyle version, I can see there are three violations in `InstancesList.java`, `BlankFormsFragment.java` and `FilledFormsFragment.java`. Fix them and get the PR passed.

thank u @huangyz0918 , i will just do it

config/quality.gradle
```diff
@@ -105,3 +105,8 @@ android {
         xmlOutput file("$reportsDir/lint/lint-result.xml")
     }
 }
+
```
huangyz0918: Looks good, but don't add this piece of code since it has been merged into master.

iamvasuh: yes sure

iamvasuh:
i have created a new one.
its all messed up here and files have not been updated

PR #207, opened 2019-03-12
Hotspot name is changing immediately now

### LINKED ISSUE #153: Hotspot name not changing immediately
Android v8.1.0, device used Asus max m1

#### Problem description
Hotspot name not changing immediately

#### Step to reproduce
1.open settings
2.click on Hotspot SSIS
3 Change name
4 Name is not changing

#### Expected behavior
name should be change immediately

Chromicle:
Closes #153 

Tested on my phone Samsung J8 is working fine as expected

I gave the else statement in that I made using the setsummary so then it is updating immediately

By updating the name immediately the user knows he updated the name instead of he has to come back

#### GIF 

immediately


Chromicle:
@lakshyagupta21 Can I also add the Toast message that like `Hotspot name is updated`

lakshyagupta21:
I don't think it's required because anyways user can see the updated hotspot name at the top item.

PR #214, opened 2019-03-16
Runtime permission

### LINKED ISSUE #119: TargeSdkVersion should be 26 or higher
Currently, the targetSdkVersion is 22 which is no longer supported.

Here's the lint error.

screenshot_2

* I think we should update it to 28.
* After updating it to 28, we will also have to update the compileSdkVersion to 28 as well.
@lakshyagupta21 @shobhitagarwal1612 What you think?

udhay24:
Closes #119 

runtime time permissions for storage, location, camera are asked depending on the SDK version
and fallback scenario if the user denied the permission is also provided
asked permissions at the required time. no specific approach used
for end users running on android M and above needs to provide the required run time permission. have to modularise few code so that they can run once the permission is accepted.
- [x] run `./gradlew lint findbugs checkstyle pmd` and confirmed all checks still pass OR confirm CircleCI build passes


shobhitagarwal1612:
This PR should only target at increasing sdk version to 26 and adding runtime permissions. Please revert other changes to reduce risks

udhay24:
I will close this PR since it contains other commits/changes. refer to PR #218 for the target sdk issue #119

PR #217, opened 2019-03-19
Added icons to all forms and settings

### LINKED ISSUE #212: Add Icons to forms and settings
#### Feature description
Currently, there are no icons to forms and settings so by adding the icons ideally also communicate the core essence of the feature

Not a problem new feature

#### Expected behavior
Good icons with vectors drawable resource files, the icon should match with feature and icon

#### Other things
while making the icons make the "Storage Location"  add the new text view and add ODK destination directory to in it

Chromicle:
Closes #212 
Tested on my phone it is working fine as expected

#### GIF
icons

for each icon added the drawable files and then attached to each particular thing

 by adding the icons ideally also communicate the core essence of the feature so it will give the good user experience

- [x] run `./gradlew check code` and confirmed all checks still pass OR confirm CircleCI build passes


shobhitagarwal1612:
+1 for the icons in the settings screen

For the forms icon, have a look at the latest release of ODK Collect. It's UI/UX was recently updated and this project should use the same iconography since the forms are shared between both

Chromicle:
Did I add the blue icon only or else will I change the color to black

shobhitagarwal1612:
> Did I add the blue icon only or else will I change the color to black

Sorry but I didn't get your question. Can you please rephrase this?

Chromicle:
There are 2 type of icons are there for filled forms and blank forms know, did I make same colour or not

Chromicle:
@shobhitagarwal1612 I updated according to odk collect I updated the PR template have a look once

lakshyagupta21:
@Chromicle It looks like that there is no left margin added for the forms icon.

Chromicle:
> @Chromicle Also if you've used the icons XML from any open source project then please mention in an open_source_licenses file (If it's not there);

Yeah I specified now

Chromicle:
@shobhitagarwal1612 please review now I made some changes

shobhitagarwal1612:
Thanks for this addition @Chromicle!

PR #221, opened 2019-03-21
password hidden while entering

### LINKED ISSUE #216: Password is not hiding while entering the password
Android Samsung j8

#### Problem description
While we entering the password the text should i.e it will show dots after entering each character

go to settings and enable set password and tap on set hotspot password and enter anything observe the behavior

#### Expected behavior
while entering the password it should hide after each character

rob729:
Closes #216 

Tested the app on my device(Lenovo P2) and the issue is solved now.

This is best possible solution as it gets the work done with minimum addition in the lines of code.

The passwords remains hidden after user has entered it.

### DIFF
--- a/skunkworks_crow/src/main/res/xml/preferences_menu.xml
+++ b/skunkworks_crow/src/main/res/xml/preferences_menu.xml
@@ -13,6 +13,7 @@
             android:summary="********"
             android:title="@string/title_hotspot_password"
             android:key="hotspot_password"
+            android:inputType="textPassword"
             android:enabled="false" />
 
         <CheckBoxPreference


rob729:
@shobhitagarwal1612 @lakshyagupta21 Please review this PR.

shobhitagarwal1612:
Now that the password is hidden, there should be a button to check the entered password for users to be able to check when needed

rob729:
@shobhitagarwal1612 There is no option to allow the toggling of visibility of text in EditTextPreference.
So what should I do now?

shobhitagarwal1612:
Have a look at TextInputEditText widget

rob729:
I don't think it will work with Shared Preferences.

shobhitagarwal1612:
You can create a custom dialog with the above widget and then save the password on OK button press.

Chromicle:
> @shobhitagarwal1612 There is no option to allow the toggling of visibility of text in EditTextPreference.
> So what should I do now?

Yeah  Toggling is not working but you can make checkbox like show password below the input of password

shobhitagarwal1612:
@Chromicle Falling back to checkbox would be ok if there was no other option. But the above alternate I mentioned is the standard way to implement it. So, it makes sense to try that out first

Chromicle:
> @Chromicle Falling back to checkbox would be ok if there was no other option. But the above alternate I mentioned is the standard way to implement it. So, it makes sense to try that out first

Yeah making cheak box instead of toogle button is not at all good but when I tried to put the toogle button it does not give any warning but the toggle but not appearing so that is what the reason I suggested for cheak box 
@rob729 while you are doing this PR for odk directory make onother preference content like storage instead of Hostspot settings

lakshyagupta21:
> I tried to put the toogle button it does not give any warning but the toggle but not appearing so that is what the reason I suggested for cheak box

I didn't understand what you're trying to say. Can you rephrase it and explain better.


> @rob729 while you are doing this PR for odk directory make onother preference content like storage instead of Hostspot settings

And the scope of this issue is just to hide character on each new character entry in the field. Lets create separate issue for this and resolve in new PR. 
@rob729 Thanks for taking up this issue, I would suggest you to try `TextInputEditText` as suggested by @shobhitagarwal1612 , if you face any difficulty in implementing it let us know.

lakshyagupta21:
@rob729 You still working on this?

rob729:
@lakshyagupta21 No

lakshyagupta21:
Closing this for now, let someone else take this up then.

PR #226, opened 2019-03-25
[#223] adds checks onBackPress()

### LINKED ISSUE #223: Stop dialog is still shown if hotspot is not enabled.
Android v7.1

#### Problem description
If user doesn't trigger hotspot from the settings screen, then it still asks  `stop sending` when the user taps back button.

1. Tap on 'Send' button.
2. Select forms to send.
3. Tap on 'Send'
4. Tap on 'Settings'
5. Don't enable from the settings screen opened (press back button)
5. A screen shows message 'Problem in creating hotspot'
6. Press the back button
7. Dialog is shown to notify a user to stop sending the forms
#### Expected behaviour
No dialog should be shown to the user if hotspot not started.
#### Other information 

nishontan:
Closes #223 

   1.  Tap on 'Send' button.
   2.  Select forms to send.
   3.  Tap on 'Send'
   4.  Tap on 'Settings'
   5.   Don't enable from the settings screen opened (press back button)
   6.   A screen shows message 'Problem in creating hotspot'
   7.   Press the back button
   8.  Dialog is not shown to notify a user to stop sending the forms

It satisfies the stated expected behaviour in the issue. Maybe wrap the codeblock
```
stopHotspot();
finish();
````
in a method named `exitSafely()`
  

- User pressed back button,``stopHotspotAlertDialog()`` is only show if hotspot is actually active i.e. ``isHotspotRunning`` is set to true 

### CODE REVIEW

**Comment by nishontan** [comment]
Why did the checks fail 🤔

**Comment by shobhitagarwal1612** [comment]
Try rerunning the circle CI task. It seems to be the problem with CircleCI

**Comment by shobhitagarwal1612** [comment]
Please fix merge conflicts

File: skunkworks_crow/src/main/java/org/odk/share/activities/SendActivity.java [hunk 1]
```diff
@@ -130,6 +130,12 @@ protected void onCreate(Bundle savedInstanceState) {
 
     @Override
     public void onBackPressed() {
+        if (!isHotspotRunning) {
+            stopHotspot();
```
shobhitagarwal1612: If the hotspot is not enabled then why are we calling `stopHotspot`?

skunkworks_crow/src/main/java/org/odk/share/activities/SendActivity.java
```diff
@@ -130,6 +130,12 @@ protected void onCreate(Bundle savedInstanceState) {
...
+            finish();
+            return;
```
shobhitagarwal1612: Instead of returning here, use if-else block

skunkworks_crow/src/main/java/org/odk/share/activities/SendActivity.java
nishontan: I blindly copy pasted positive button callback from `stopHotspotAlertDialog`
I agree and understand that block is unnecessary now.

PR #227, opened 2019-03-25
Add support for Android 9

### LINKED ISSUE #191: Add support to send forms on Android 9
Android 9, Moto G5 Plus

#### Problem description
Unable to send blank/filled forms on Android 9. The screen gets stuck on `Waiting for the hotspot to connect` 

Just select any blank or filled form and try to send on an android 9 device

#### Expected behavior
It should work as expected

#### Other information 

shobhitagarwal1612:
Closes #191

Tested by sending forms multiple times on MotoG5 Plus (Android 9)

Not the best solution but this is the best I could come up with. Please suggest if you have any other approaches. So the current approach goes like this:
 - Removed `startLocalOnlyHotspot` as server socket wasn't reachable with that way. It works perfectly to start/stop hotspot but the ultimate goal wasn't being served. I searched a lot but wasn't able to find anything solid as to why the sockets were unreachable. Although, the host was pingable (tested by `ping <host_ip>` from the shell of the receiver device)
 - Only system apps have permission to read/write wifi hotspot configurations. So, we have to rely on manual entry of ssid and password which happens on the first launch of the app for Android 9 devices.
 - Since users might not change their default hotspot's name, I had to remove the filter which only shows devices running the same app. We can explore [NSD](https://developer.android.com/training/connect-devices-wirelessly/nsd) for that purpose in the future.
 - Also fixes a bug when play services are not available

 Need feedback for this approach. Might still contain bugs but at least it works now.


shobhitagarwal1612:
I've requested a review from people I remember might have mentioned that they have an Android 9 device to try this PR out

udhay24:
@shobhitagarwal1612 i don't have a android 9 device now but i can get my hands on one by monday

lakshyagupta21:
@shobhitagarwal1612 Please fix the conflicts.

shobhitagarwal1612:
Thanks for testing it out @Chromicle and I'm glad that it worked for you. Could you please mention which all devices you used for testing this PR? It would be nice to keep a track of devices and android versions

Chromicle:
Samsung J8 of Android 8.0.0 and samsung note 3 of Android 9 these all devices I tested it is working fine

shobhitagarwal1612:
> @shobhitagarwal1612 Please fix the conflicts.

@lakshyagupta21 Done

udhay24:
ezgif com-video-to-gif

@shobhitagarwal1612 i just tested it in samsung j7(2016) - android 8.1. I noticed that AlertDialog keeps popping one after the another until the app becomes un-responsive

shobhitagarwal1612:
I found another bug, not the one you have mentioned @udhay24. Also, I tested on Android 9 as I don't have an Android 8 device

Chromicle:
@udhay24 I just now checked I did not find that type of behavior in android 8.0.0 
@shobhitagarwal1612 I also found the bug is when the hotspot is on initially now also it is asking to on the hotspot

shobhitagarwal1612:
I didn't find a way to get the status of the hotspot in Android 9. Hence the alert dialog. 

My intention behind this PR was to give a push to this GSoC project and get the base ready i.e. sending/receiving so that you guys can focus on new features and overall stability. All these issues can be open separately and fixed along during the summer

udhay24:
> ezgif com-video-to-gif
> 
> @shobhitagarwal1612 i just tested it in samsung j7(2016) - android 8.1. I noticed that AlertDialog keeps popping one after the another until the app becomes un-responsive

@shobhitagarwal1612 i forgot to mention that the location was already enabled but the app was still asking to enable the location.

shobhitagarwal1612:
@udhay24 I was able to reproduce the issue. Please review again

skunkworks_crow/src/main/java/org/odk/share/activities/ReceiverActivity.java
```diff
@@ -397,19 +399,25 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
+                    ip = (String) obj.get(IP);
+                    wifiNetworkSSID = (String) obj.get(SSID);
                     port = (Integer) obj.get(PORT);
                     isProtected = (boolean) obj.get(PROTECTED);
                     if (isProtected) {
                         passwordScanned = (String) obj.get(PASSWORD);
                     }
 
-                    Timber.d("Scanned results " + ssidScanned + " " + port + " " + isProtected + " " + passwordScanned);
+                    Timber.d("Scanned results " + wifiNetworkSSID + " " + port + " " + isProtected + " " + passwordScanned);
```
huangyz0918: I suggest removing all the test log if they are only used for debugging, or just comment them.

skunkworks_crow/src/main/java/org/odk/share/activities/ReceiverActivity.java
```diff
@@ -397,19 +399,25 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
                 // request was canceled...
                 Timber.i("QR code scanning cancelled");
             } else {
-                Timber.d("RESULT " + result);
```
huangyz0918: The same as above.

skunkworks_crow/src/main/java/org/odk/share/activities/ReceiverActivity.java
```diff
@@ -491,12 +495,12 @@ private void wifiListAvailable(ArrayList<WifiNetworkInfo> list) {
...
             isQRCodeScanned = false;
             for (WifiNetworkInfo info : list) {
-                if (info.getSsid().equals(ssidScanned)) {
+                if (info.getSsid().equals(wifiNetworkSSID)) {
                     if (info.getState() != NetworkInfo.DetailedState.CONNECTED) {
                         Toast.makeText(this, "attempting connection", Toast.LENGTH_SHORT).show();
-                        connectToNetwork(info.getSecurityType(), ssidScanned, passwordScanned);
+                        connectToNetwork(info.getSecurityType(), wifiNetworkSSID, passwordScanned);
                     } else {
-                        Toast.makeText(this, "already connected to " + ssidScanned, Toast.LENGTH_SHORT).show();
```
huangyz0918: Can we extract the toasts into something like `ToastUtils.java`? You can refer to [Collect's](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/utilities/ToastUtils.java). Something for showing dialogs can be refactor as well.

skunkworks_crow/src/main/java/org/odk/share/activities/SendActivity.java
```diff
@@ -106,11 +102,10 @@ protected void onCreate(Bundle savedInstanceState) {
             formIds = getIntent().getLongArrayExtra(FORM_IDS);
         }
 
-
         port = SocketUtils.getPort();
 
         if (port == -1) {
```
huangyz0918: `-1` can be replaced by a static variable with meaningful name, like `NO_AVAILABLE_PORT_CODE`?

skunkworks_crow/src/main/java/org/odk/share/activities/SendActivity.java
```diff
@@ -168,141 +164,102 @@ private void startHotspot() {
...
-                finish();
-            }
-        });
+        new AlertDialog.Builder(this)
+                .setMessage(R.string.hotspot_settings_dialog)
+                .setPositiveButton(getString(R.string.settings), (dialog, which) -> {
+                    isHotspotInitiated = true;
+
+                    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+                        // do nothing
```
huangyz0918: `//TODO: ` will be helpful.

skunkworks_crow/src/main/java/org/odk/share/activities/SendActivity.java
```diff
@@ -423,49 +360,36 @@ protected Dialog onCreateDialog(int id) {
...
-                        stopHotspot();
-                        finish();
-                        break;
-                }
-            }
-        };
-        alertDialog.setCancelable(false);
-        alertDialog.setButton(DialogInterface.BUTTON_POSITIVE, getString(R.string.ok), quitListener);
-        alertDialog.show();
+        new AlertDialog.Builder(this)
```
huangyz0918: So many similar dialogs, can we building something like `DailogFactory.java` which has different static methods can return dialogs?

skunkworks_crow/src/main/java/org/odk/share/activities/SendActivity.java
```diff
@@ -487,23 +411,10 @@ public void onRequestPermissionsResult(int requestCode, @NonNull String[] permis
...
 
     private void toggleHotspot() {
-
         if (!isHotspotInitiated) {
-            isHotspotInitiated = true;
             startHotspot();
-        }
-
-        if (openSettings) {
+        } else if (openSettings) {
```
huangyz0918: `openSettings` is a bad name for a boolean value.

skunkworks_crow/src/main/java/org/odk/share/network/WifiHotspotManager.java
```diff
@@ -0,0 +1,81 @@
...
+            // show system alert dialog to enable wifi tethering
+            throw new TriggerHotspotException();
+        } else {
+            // enable wifi tethering
+            wifiHotspot.enableHotspot(sharedPreferencesHelper.getHotspotName());
+        }
+    }
+
+    public void stopHotspot() {
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.N) {
```
huangyz0918: For those version checks, we can build a boolean variable like `isApiHigherThanO = ...`, can call `isApiHigherThanO` or `!isApiHigherThanO` if necessary, which can simplify the code and make them better to read.

shobhitagarwal1612:
@lakshyagupta21 What are your thoughts about this PR? Do you think this would be a good solution to some of the mention wifi hotspot issues?
If yes, then would it be a feasible and good task for @huangyz0918 to finish off the remaining work in this PR by adjusting some of the non-crucial tasks in his timeline after he is done with https://github.com/opendatakit/skunkworks-crow/pull/260?

lakshyagupta21:
> @lakshyagupta21 What are your thoughts about this PR? Do you think this would be a good solution to some of the mention wifi hotspot issues?

Even I think that should resolve some of the issues. yeah it would be a good idea to let @huangyz0918 pick this up, if he needs some help with the Android 9 testing in case he doesn't have one then we can hop in.

huangyz0918:
> yeah it would be a good idea to let @huangyz0918 pick this up, if he needs some help with the Android 9 testing in case he doesn't have one then we can hop in.

Yeah I'd like to, but I don't have two android 9 devices currently, so I may have some trouble testing the changes. Can you help me to test the PR #260 using android 9 first? @lakshyagupta21 I want to know what kinds of issue in android 9 (I have used the android 8.0.0, 8.0.1 and 6.0.1 tested with success), for you guys have android 9 devices @Chromicle @udhay24 @iadeelzafar , I think it will be very nice to have a quick test with that, thanks so much!!

shobhitagarwal1612:
You don't need to have both android 9 devices at the same time. The same device can be used for validating sending/receiving one at a time.

huangyz0918:
> You don't need to have both android 9 devices at the same time. The same device can be used for validating sending/receiving one at a time.

I think I can flash my Nexus 6P to android 9 after #260 was merged.

skunkworks_crow/src/main/java/org/odk/share/activities/SendActivity.java
```diff
@@ -88,9 +87,6 @@
     private long[] formIds;
     private int mode;
 
-    private WifiManager.LocalOnlyHotspotReservation hotspotReservation;
```
lakshyagupta21: @shobhitagarwal1612 Is there any specific reason for removing this class and its API usage? I think Google has introduced this Class and some APIs so that hotspot connection can be made and two nearby devices can share the data locally without sharing the user's mobile data.

PR #230, opened 2019-03-29
[#216] replaced edittextpref with textinputlayout dialog

### LINKED ISSUE #216: Password is not hiding while entering the password

nishontan:
Closes #216 

- Opened settings screen
- Checked if toggle worked
- Checked if changed password was saved
- Checked if default password was loaded, when setting password for the first time

password-toggle

- Easiest method of achieving password toggle
- However, using a checkbox to toggle was considered but was quickly ruled out after following the discussion on #221 

- User can now toggle password in input area

### DIFF
--- a/skunkworks_crow/src/main/java/org/odk/share/preferences/SettingsPreference.java
+++ b/skunkworks_crow/src/main/java/org/odk/share/preferences/SettingsPreference.java
@@ -157,7 +157,5 @@ public class SettingsPreference extends PreferenceActivity {
         AlertDialog alertDialog = builder.create();
         alertDialog.show();
     }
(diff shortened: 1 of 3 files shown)


nishontan:
wifi-dialog

There is a implementation of password dialog on ``WifiActivity`` that uses a checkbox to toggle 
If this PR gets merge I will create a new issue to replace that dialog

PR #243, opened 2019-04-08
When we pressed back in set Hotspot it is coming back and updated key board

### LINKED ISSUE #241: When back pressed on set hotspot password it is not going to back and does not open soft keyboard automatically
Android Samsung j8

#### Problem description
when we pressed back button instead of cancel button the alert dialog should be canceled but it is not coming back 
when the alert dialogue is opened the soft keyboard should open automatically but it does not happen

Go to the settings screen
press the Set Hotspot password
instead of cancel button use the back button  to cancel alert dialogue
see the behavior 

#### Expected behavior
when we pressed back it should be canceled and we will get into the settings to screen only
when we clicked on set hot spot password soft keyboard should open automatically

Chromicle:
Closes #241 

Checked whether is it coming back to settings activity or not
Checked whether the soft keyboard is coming automatically or not
checked whether any new warnings are produced or not
it is working fine as expected

I think instead of adding method `OnBackPressed`  by adding `setCancelable(true)` it will be good
For soft keyboard by the support of `Windows manager` I used `SOFT_INPUT_STATE_ALWAYS_VISIBLE` so it is automatically visible

If the user wants to come back instead of pressing the back button he can directly come back by pressing the back button
By getting automatically soft keyboard user no need to touch and activate keyboard 


Chromicle:
@shobhitagarwal1612 @lakshyagupta21 I added another small feature(about soft keyboard) in this same PR I think no need of onother issue for that So I added in this PR only

PR #246, opened 2019-04-09
Fixed App crash in open source licenses

### LINKED ISSUE #245: App crashes on clicking Open source licenses
Android Samsung J8

#### Problem description
App crashes on clicking open source licenses 
App should not be crash  as app crash is not at all good behavior to any user 

Click on about
And click on Open source licenses
observe the behavior

#### Expected behavior
App should not be crash and should open web view of open source licenses

Chromicle:
Closes #245 

Checked whether App is crashing or not
checked whether web view of open source licenses are opening or not
And checked any new warnings are producing  

As by using `android.support.v7.widget.Toolbar` so the layout is inflating so I changed to  `androidx.appcompat.widget.Toolbar ` so by using the support of android RX so I fixed that

By solving the app crash on clicking so that a user can see the web view of open source licenses
 

### DIFF
--- a/skunkworks_crow/src/main/res/layout/activity_web_view.xml
+++ b/skunkworks_crow/src/main/res/layout/activity_web_view.xml
@@ -1,32 +1,30 @@
-<RelativeLayout
-    xmlns:android="http://schemas.android.com/apk/res/android"
+<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
-    android:layout_marginBottom="0dp"
     android:layout_marginLeft="0dp"
-    android:layout_marginRight="0dp"
     android:layout_marginTop="0dp"
+    android:layout_marginRight="0dp"
+    android:layout_marginBottom="0dp"
     android:orientation="vertical"
     android:padding="0dp">
 
-    <android.support.v7.widget.Toolbar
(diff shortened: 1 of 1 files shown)


Chromicle:
@shobhitagarwal1612 @lakshyagupta21 can you please review it

skunkworks_crow/src/main/res/layout/activity_web_view.xml
```diff
@@ -1,32 +1,30 @@
...
-    android:layout_marginRight="0dp"
     android:layout_marginTop="0dp"
+    android:layout_marginRight="0dp"
+    android:layout_marginBottom="0dp"
     android:orientation="vertical"
     android:padding="0dp">
 
-    <android.support.v7.widget.Toolbar
-        xmlns:android="http://schemas.android.com/apk/res/android"
+    <androidx.appcompat.widget.Toolbar xmlns:android="http://schemas.android.com/apk/res/android"
```
shobhitagarwal1612: The namespace `android` is already defined in the root element. This one can be removed

Chromicle: I didn't add that, anyway that is no need

shobhitagarwal1612: I must have missed it while it was added in one of the earlier PRs. Apologies. And thanks for fixing it

PR #251, opened 2019-04-29
feat: Made Directory Picker instead of edittext

### LINKED ISSUE #250: Instead of editText for Destination Directory make directory picker from storage
Android  Samsung J8

#### Problem description
Currently for `ODK Destination Directory` there is an edit text for file path by that user may type wrong spelling then the File path does not exist
Instead of that making directory picker so user can directly select the directory from his storage

#### Steps to reproduce Behaviour
Go to the settings 
Click on `ODK Destination Directory` 
As to avoid the typo problem make directory picker

#### Expected behavior
Make Directory picker to select any file directly instead of typing in edit text

Chromicle:
Closes #250 

Checked whether selected Directory is taking as a summary and file path value
Checked all other preferences whether they are working properly or not
Checked in some phones(Android 7,8,9) whether the intent is working properly or not
Checked whether it is taking sdCard or not along with internal storage

I made Intent `Intent.ACTION_OPEN_DOCUMENT_TREE` so by clicking on that we can see all files in a device  
By that, I added another method `onActivityResult` so we can get the file path through `(data.getData()).getPath()` without any typos

Mainly, It will avoid the typos regarding Directory names and user can directly select the Directory easily

#### GIF
ezgif com-video-to-gif


skunkworks_crow/src/main/java/org/odk/share/preferences/SettingsPreference.java
```diff
@@ -159,4 +155,25 @@ private void showPasswordDialog() {
...
+        Intent i = new Intent(Intent.ACTION_OPEN_DOCUMENT_TREE);
+        i.addCategory(Intent.CATEGORY_DEFAULT);
+        startActivityForResult(Intent.createChooser(i, getString(R.string.choose_directory)), DIRECTORY_REQUEST_CODE);
+    }
+
+    public void onActivityResult(int requestCode, int resultCode, Intent data) {
+        switch (requestCode) {
+            case DIRECTORY_REQUEST_CODE:
+                try {
+                    String filePath = (data.getData()).getPath();
```
lakshyagupta21: `(data.getData()).getPath()` can be written as `data.getData().getPath()`

skunkworks_crow/src/main/java/org/odk/share/preferences/SettingsPreference.java
```diff
@@ -159,4 +155,25 @@ private void showPasswordDialog() {
...
+                    filePath = filePath + getString(R.string.directory_odk);
```
lakshyagupta21: `filePath = filePath + getString(R.string.directory_odk);` can be written as  `filePath = filePath + getString(R.string.directory_odk);`

skunkworks_crow/src/main/java/org/odk/share/preferences/SettingsPreference.java
```diff
@@ -159,4 +155,25 @@ private void showPasswordDialog() {
         AlertDialog alertDialog = builder.create();
         alertDialog.show();
     }
+
+    public void chooseDirectory() {
```
lakshyagupta21: @Chromicle Have you tried handling when there is no intent to handle directory picker?

lakshyagupta21:
Also resolve conflicts.

skunkworks_crow/src/main/java/org/odk/share/preferences/SettingsPreference.java
Chromicle: Yeah, I tried but it does not give better results alike as putting intent 
But I checked in Android 6,7,8,9 phones whether the intent is working or not it is perfectly working

lakshyagupta21: I tried this in Android 4.4 version and I was not able to select any directory. Can you please verify or can you think of any fallback mechanism?

Chromicle: Since there is no "default file manager" in KitKat it is not working if the user is installed a particular 3rd party file manager then it may works

lakshyagupta21: @Chromicle If we are adding some feature then we should be confident enough that it won't break anything which was working previously.  So adding this feature will not work as well as won't allow user to manually write the directory path. So I would suggest you rather than printing log or toast it would be beneficial to have some fallback mechanism for some devices in which directory picker is not available.
One more suggestion I would like to give that whenever you're doing some kind of testing then always try to test on less than 5.

lakshyagupta21:
@Chromicle Are you still working on this, because I see that few of the above issues are still not resolved related to lower Android devices

Chromicle:
> @Chromicle Are you still working on this, because I see that few of the above issues are still not resolved related to lower Android devices
@lakshyagupta21 I resolved in the way that for android version less than KitKat they will get the edit text and android versions greater than that will get the directory picker
I pick this approach because we can not the data from the path of lesser android versions

lakshyagupta21:
@Chromicle I would recommend you to run `./gradlew checkCode` on every commit you push because sometimes we miss something due to which it fails and also check the checkboxes in the PR description that you've done both the things mentioned there.

> #### Before submitting this PR, please make sure you have:
> * [ ]  run `./gradlew checkCode` and confirmed all checks still pass OR confirm CircleCI build passes
> * [ ]  verified that any code or assets from external sources are properly credited in comments and/or in the [about file](https://github.com/opendatakit/share/blob/master/share_app/src/main/assets/open_source_licenses.html).

lakshyagupta21:
@Chromicle Please update your PR template.

Chromicle:
> @Chromicle Please update your PR template.

Yeah I updated can you please check it

skunkworks_crow/src/main/java/org/odk/share/views/ui/settings/SettingsActivity.java
```diff
@@ -29,10 +33,12 @@
 
 public class SettingsActivity extends PreferenceActivity {
 
+    private static final int DIRECTORY_REQUEST_CODE = 9999;
     EditTextPreference hotspotNamePreference;
     Preference hotspotPasswordPreference;
-    CheckBoxPreference passwordRequirePreference;
-    EditTextPreference odkDestinationDirPreference;
+    SwitchPreference passwordRequirePreference;
+    Preference odkDestinationDirPreferenceDirectoryPicker;
```
lakshyagupta21: Creating two different preferences is fine, but the values that we get from these two different preferences on different versions stored separately. Ideally, whatever value that we receive from these UI Objects should be saved to the same preference.

skunkworks_crow/src/main/java/org/odk/share/tasks/DownloadJob.java
```diff
@@ -362,9 +363,18 @@ private String getInstancesPath() {
 
     private String getOdkDestinationDir() {
         SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(getContext().getApplicationContext());
+        String directoryPath = null;
 
-        return prefs.getString(PreferenceKeys.KEY_ODK_DESTINATION_DIR,
-                getContext().getString(R.string.default_odk_destination_dir));
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
```
lakshyagupta21: Did you check this behavior on any of the lower-end devices, like Android 4.4(Kitkat) or JellyBean (4.3) or Android 5.1 (Lollipop)?

lakshyagupta21:
@Chromicle due to some recent commits, now this PR again have some merge conflicts can you resolve them, and do a dev testing like transfer a form from one device to another and check if you're able to see the entries in the selected directory or not.

skunkworks_crow/src/main/java/org/odk/share/views/ui/settings/SettingsActivity.java
Chromicle: I tried that but I am getting preference cast exception

skunkworks_crow/src/main/java/org/odk/share/tasks/DownloadJob.java
Chromicle: Yeah I checked in android 5 it is working perfectly with edit text

skunkworks_crow/src/main/java/org/odk/share/views/ui/settings/SettingsActivity.java
lakshyagupta21: I'll give it a shot then will suggest the solution

skunkworks_crow/src/main/java/org/odk/share/tasks/DownloadJob.java
lakshyagupta21: I would suggest you to test on Android 4.4, I think this will fail on that device.

skunkworks_crow/src/main/java/org/odk/share/views/ui/settings/SettingsActivity.java
Chromicle: Yeah than you

skunkworks_crow/src/main/java/org/odk/share/tasks/DownloadJob.java
Chromicle: Sorry I dont have Android 4.4 device I checked in 5,6,7,8,9 it is working perfectly

Chromicle: But I send my branch to onother guy he tested and he said that it is working with edit text perfectly as before

PR #260, opened 2019-06-05
[MRG] Bluetooth Feature for Form Exchanging

### LINKED ISSUE #253: [Feature] Add the bluetooth development framework
#### Problem description
We are going to build bluetooth methods. Should we using the native bluetooth server/client in Android SDK or using other open source wrapped bluetooth lib? 

A suggestion is [FastBle](https://github.com/Jasonchenlijian/FastBle), which uses simple ways to filter, scan, connect, read ,write, notify, readRssi, setMTU, and multiConnection using bluetooth. I was a user of this library and it works very good in my other applications. 

#### Expected behavior
Build native bluetooth server/client utils or adding a third-party bluetooth library.

#### Other information 
@opendatakit-bot claim

huangyz0918:
Closes #253 #265 

I have tested it using my Huawei Nexus 6P (Android 8.0) and Meizu M5 (Android 6.0).

Now, we are going to build a bluetooth transfer feature. Using bluetooth, we can share form instances between two devices. For bluetooth BLE and bluetooth classic, I choose bluetooth classic because bluetooth classic is compatible with more devices (BLE can only supports device with Android Version >= 4.3).

I'm going to introduce the basic bluetooth tools for ODK Skunkworks-Crow in the PR, The basic tools can be divides into two parts, the bluetooth client and bluetooth server. For data  transmission, we need one device act as a bluetooth client and another as a server. If one is not ready, the bluetooth transmission will failed. So I want to build an alert dialog to ask which method our user would like to use, the Wifi Hotspot? or the Bluetooth? Only two device choose the same approach, the connection can be established. 

For building this new feature, we have several tasks, 

- [x] build basic bluetooth transfer client method. 
- [x] build basic bluetooth transfer server method.
- [x] build a new `BluetoothActivity.java` for showing scanning results and device list.
- [x] build data transfer methods (short messages, files).
- [x] rename files to make them more universal (e.g, `HpSenderActivity.java` and `BtSenderActivity.java`)
- [x] build the basic alert dialog to ask user to choose a transfer method.
- [x] build the sending/receiving forms (instance) feature.
- [x] improve the UI/UX progress.

It needs tests with new features and make sure the old features work well.


huangyz0918:
Now, I finished the bluetooth tools for file sending and receiving, I made a demo for sending a short message using bluetooth, you can see this [video demo](https://www.dropbox.com/s/1uj9lfw97bsjri4/demo.mp4?dl=0).

I created a `bluetooth` package in the root of this project, 
- `BluetoothBasic.java`: Basic methods for bluetooth connecting/sending/loop reading.
- `BluetoothReceiver.java`: `BroadcastReceiver` for system bluetooth services.
- `BluetoothClient.java`: extend from `BluetoothBasic.java`, for data sending.
- `BluetoothServer.java`: extend from `BluetoothBasic.java`, for data receiving.
- `BluetoothUtils.java`: common methods for bluetooth status checking, etc.

huangyz0918:
@lakshyagupta21 @shobhitagarwal1612 I think it's ready for a review.

skunkworks_crow/src/main/java/org/odk/share/views/ui/send/fragment/FilledFormsFragment.java
```diff
@@ -150,13 +155,40 @@ private void setEmptyViewVisibility(int len) {
 
     @OnClick(R.id.send_button)
     public void send() {
-        Intent intent = new Intent(getActivity(), SendActivity.class);
+        Intent hotspotIntent = new Intent(getActivity(), HpSenderActivity.class);
```
lakshyagupta21: Create intent in the switch block itself and only one intent is required because there can only be one option either hotspot or bluetooth.

skunkworks_crow/src/main/java/org/odk/share/views/ui/send/fragment/FilledFormsFragment.java
```diff
@@ -45,6 +49,7 @@
     public static final String INSTANCE_IDS = "instance_ids";
     private static final String INSTANCE_LIST_ACTIVITY_SORTING_ORDER = "instanceListActivitySortingOrder";
     private static final int INSTANCE_LOADER = 1;
+    private static final String[] TRANSFER_OPTIONS = {"via bluetooth", "via hotspot"};
```
lakshyagupta21: Lets create or resuse the string resources and `via` can be removed from the string.

skunkworks_crow/src/main/java/org/odk/share/views/ui/send/fragment/BlankFormsFragment.java
```diff
@@ -156,13 +161,39 @@ private void setEmptyViewVisibility(int len) {
 
     @OnClick(R.id.send_button)
     public void send() {
-        Intent intent = new Intent(getActivity(), SendActivity.class);
+        Intent hotspotIntent = new Intent(getActivity(), HpSenderActivity.class);
```
lakshyagupta21: Same here as mentioned in FilledFormsFragment.

skunkworks_crow/src/main/java/org/odk/share/views/ui/main/MainActivity.java
```diff
@@ -103,10 +106,28 @@ private void setupAdapter() {
         recyclerView.setAdapter(formAdapter);
     }
 
-    @OnClick(R.id.bViewWifi)
-    public void viewWifiNetworks() {
-        Intent intent = new Intent(this, WifiActivity.class);
-        startActivity(intent);
+    @OnClick(R.id.bReceiveForms)
+    public void chooseReceivingMethods() {
+        Intent hotspotIntent = new Intent(this, HpReceiverActivity.class);
```
lakshyagupta21: Same as mentioned in FilledFormsFragment.

skunkworks_crow/src/main/java/org/odk/share/views/ui/main/MainActivity.java
```diff
@@ -103,10 +106,28 @@ private void setupAdapter() {
...
+        Intent bluetoothIntent = new Intent(this, BtReceiverActivity.class);
+
+        // choose different sending methods, TODO: needs refactor if we have a better UI/UX.
+        android.app.AlertDialog alertDialog = new android.app.AlertDialog.Builder(this)
+                .setTitle("Receive Options")
+                .setIcon(R.drawable.ic_help_outline)
+                .setItems(TRANSFER_OPTIONS, (DialogInterface dialog, int which) -> {
+                    switch (which) {
+                        case 0:
+                            startActivity(bluetoothIntent);
```
lakshyagupta21: create intent here and call 
`startActivity(intent);`
`finish();`
after switch block as this is a common statement for both the options of switch block

skunkworks_crow/src/main/java/org/odk/share/views/ui/main/MainActivity.java
```diff
@@ -52,12 +54,13 @@
     private static final String COLLECT_PACKAGE = "org.odk.collect.android";
     private static final int STORAGE_PERMISSION_REQUEST_CODE = 101;
     private static final int FORM_LOADER = 2;
+    private static final String[] TRANSFER_OPTIONS = {"via bluetooth", "via hotspot"};
```
lakshyagupta21: Same as commented below, use string resources.

skunkworks_crow/src/main/java/org/odk/share/views/ui/instance/fragment/ReviewedInstancesFragment.java
```diff
@@ -224,7 +224,7 @@ public void sendForms() {
                 instanceIds.add(transferInstance.getInstanceId());
             }
         }
-        Intent intent = new Intent(getContext(), SendActivity.class);
+        Intent intent = new Intent(getContext(), HpSenderActivity.class);
```
lakshyagupta21: There should be a dialog asking for one of the options out of the `Bluetooth` and `Hotspot`, here it is directly triggering hotspot.

skunkworks_crow/src/main/java/org/odk/share/views/ui/instance/InstancesList.java
```diff
@@ -119,7 +119,7 @@ private void onListItemClick(View view, int position) {
 
     @OnClick(R.id.send_button)
     public void send() {
-        Intent intent = new Intent(this, SendActivity.class);
+        Intent intent = new Intent(this, HpSenderActivity.class);
```
lakshyagupta21: Same dialog should be displayed showing which mode to choose (Bluetooth or Hotspot).

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothReceiver.java
```diff
@@ -0,0 +1,99 @@
...
+        BluetoothDevice dev = intent.getParcelableExtra(BluetoothDevice.EXTRA_DEVICE);
+        if (dev != null) {
+            Timber.d(TAG + "BluetoothDevice: " + dev.getName() + ", " + dev.getAddress());
+        }
+        switch (action) {
+            case BluetoothAdapter.ACTION_STATE_CHANGED:
+                int state = intent.getIntExtra(BluetoothAdapter.EXTRA_STATE, 0);
+                Timber.d(TAG + "STATE: " + state);
+                break;
+            case BluetoothAdapter.ACTION_DISCOVERY_STARTED:
```
lakshyagupta21: You can remove the action Items which are no ops.

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothReceiver.java
```diff
@@ -0,0 +1,99 @@
...
+                break;
+            case BluetoothAdapter.ACTION_DISCOVERY_FINISHED:
+                break;
+
+            case BluetoothDevice.ACTION_FOUND:
+                short rssi = intent.getShortExtra(BluetoothDevice.EXTRA_RSSI, Short.MAX_VALUE);
+                Timber.d(TAG + "EXTRA_RSSI:" + rssi);
```
lakshyagupta21: And Timber can be used like `Timber.d("EXTRA_RSSI: %s", rssi);`
This applies to all usages.

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothUtils.java
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Basic tools for bluetooth features.
+ *
+ * @author huangyz0918 (huangyz0918@gmail.com)
+ * @since 04/06/2019
+ */
+public class BluetoothUtils {
+
+    // UUID for security bluetooth pairing.
+    public static final UUID SPP_UUID = UUID.fromString("724f00a0-795e-4272-a85a-11075e760e58");
```
lakshyagupta21: Is this required to be constant? Or it can be generated randomly at runtime.
It would be good if generate it randomly at runtime.

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothUtils.java
```diff
@@ -0,0 +1,76 @@
...
+        return isSupportBluetooth() && getBluetoothAdapter().isEnabled();
+    }
+
+    /**
+     * Disable and enable the bluetooth.
+     */
+    public static void disableBluetooth() {
+        if (isBluetoothEnabled()) {
+            new Thread(() -> {
+                getBluetoothAdapter().disable();
```
lakshyagupta21: Is this needed to be execute in thread?

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothUtils.java
```diff
@@ -0,0 +1,76 @@
...
+            }).start();
+        }
+    }
+
+    public static void enableBluetooth() {
+        if (!isBluetoothEnabled()) {
+            new Thread(() -> {
+                getBluetoothAdapter().enable();
```
lakshyagupta21: Is this required to be executed in thread.

skunkworks_crow/src/main/java/org/odk/share/services/SenderService.java
```diff
@@ -58,18 +58,54 @@ private void addUploadJobSubscription() {
...
+    /**
+     * build a bundle for transfer form data using bluetooth.
+     *
+     * @param instancesToSend instance to send.
+     * @param mode            review form mode or others.
+     */
+    private PersistableBundleCompat setupPersistableBundle(long[] instancesToSend, int mode) {
+        PersistableBundleCompat extras = new PersistableBundleCompat();
+        extras.putLongArray(UploadJob.INSTANCES, instancesToSend);
+        extras.putBoolean("isBluetooth", true);
```
lakshyagupta21: Same here, use same string constant which will be created above and pass that to identify the mode of transfer.

skunkworks_crow/src/main/java/org/odk/share/tasks/DownloadJob.java
```diff
@@ -94,26 +103,60 @@ protected Result onRunJob(@NonNull Params params) {
...
+        }
+
+        setupDataStreamsAndReceive(isBluetooth, targetMacAddress);
     }
 
-    private DownloadEvent receiveForms() {
-        Timber.d("Socket " + ip + " " + port);
+    private void setupDataStreamsAndReceive(boolean isBluetooth, String macAddress) {
+        try {
+            Timber.d("Waiting for receiver");
```
lakshyagupta21: The log should be like `Waiting for sender` because this is the `DownloadJob`

skunkworks_crow/src/main/java/org/odk/share/tasks/DownloadJob.java
```diff
@@ -94,26 +103,60 @@ protected Result onRunJob(@NonNull Params params) {
...
+            if (isBluetooth) {
+
+                BluetoothAdapter bluetoothAdapter = BluetoothAdapter.getDefaultAdapter();
+                BluetoothDevice bluetoothDevice = bluetoothAdapter.getRemoteDevice(macAddress);
```
lakshyagupta21: I think null checks would be the good idea because sometimes what if bluetooth adapter comes null or device is not able to find any connected bluetooth device with the specifies macAddress.

skunkworks_crow/src/main/java/org/odk/share/tasks/DownloadJob.java
```diff
@@ -94,26 +103,60 @@ protected Result onRunJob(@NonNull Params params) {
...
+                BluetoothSocket bluetoothSocket = bluetoothDevice.createRfcommSocketToServiceRecord(SPP_UUID);
+                if (!bluetoothSocket.isConnected()) {
+                    bluetoothSocket.connect();
```
lakshyagupta21: The thread is blocked here, and it will try to connect to the bluetooth device, if its not able to connect then it will through the `IOException` so we should fail the form transfer if exception occurs.

skunkworks_crow/src/main/java/org/odk/share/tasks/UploadJob.java
```diff
@@ -89,34 +95,62 @@ protected Result onRunJob(@NonNull Params params) {
...
+    private UploadEvent uploadInstances() {
+        try {
             // show dialog and connected
             Timber.d("Start Sending");
             processSelectedFiles(instancesToSend);
 
             // close connection
-            socket.close();
-            serverSocket.close();
+            if (socket != null) {
```
lakshyagupta21: When all the form transfer is done newly created `bluetoothSocket` also needs to closed for both sender and receiver other it may result in memory leaks.

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BluetoothListAdapter.java
```diff
@@ -0,0 +1,134 @@
...
+    public ViewHolder onCreateViewHolder(@NonNull ViewGroup parent, int viewType) {
+        View view = LayoutInflater.from(parent.getContext()).inflate(R.layout.list_item_bt_device, null);
+        return new ViewHolder(view);
+    }
+
+
+    /**
+     * rescan the bluetooth devices and update the list.
+     */
+    public void rescan() {
```
lakshyagupta21: rescan logic should not be present in adpater class same for below method `addBoundedDevices`

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtReceiverActivity.java
```diff
@@ -0,0 +1,192 @@
...
+     * application's {@link RxEventBus}
+     */
+    private Disposable addBluetoothEventSubscription() {
+        return rxEventBus.register(BluetoothEvent.class)
+                .subscribeOn(schedulerProvider.io())
+                .observeOn(schedulerProvider.androidThread())
+                .subscribe(bluetoothEvent -> {
+                    switch (bluetoothEvent.getStatus()) {
+                        case CONNECTED:
+                            progressDialog.setMessage("Connected, start downloading...");
```
lakshyagupta21: Create string resource for this or use the same that is defined for hotspot.

lakshyagupta21:
Also one thing that I noticed while scanning the devices is that it is showing all the devices which was paired before to the device but not available at the moment of scanning, so please take a look at that as well.

skunkworks_crow/src/main/AndroidManifest.xml
```diff
@@ -18,6 +18,10 @@
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
     <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
 
+    <!-- Bluetooth Related -->
```
shobhitagarwal1612: The permissions are self explanatory. No need for this comment

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothReceiver.java
```diff
@@ -0,0 +1,99 @@
...
+            case BluetoothHeadset.ACTION_CONNECTION_STATE_CHANGED:
+                Timber.i(TAG + "CONN_STATE: " + intent.getIntExtra(BluetoothHeadset.EXTRA_STATE, 0));
+                break;
+            case BluetoothA2dp.ACTION_CONNECTION_STATE_CHANGED:
+                Timber.i(TAG + "CONN_STATE: " + intent.getIntExtra(BluetoothA2dp.EXTRA_STATE, 0));
+                break;
+        }
+    }
+
+    public interface Listener {
```
shobhitagarwal1612: Please rename this interface to something less generic like `BluetoothReceiverListener`

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothReceiver.java
```diff
@@ -0,0 +1,99 @@
...
+        void foundDevice(BluetoothDevice device);
```
shobhitagarwal1612: The name of the method should display its intention. So, `onDeviceFound` would sound more accurate. Comments?

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothUtils.java
```diff
@@ -0,0 +1,76 @@
...
+    private BluetoothUtils() {
+
+    }
+
+    private static BluetoothAdapter getBluetoothAdapter() {
+        return BluetoothAdapter.getDefaultAdapter();
+    }
+
+    /**
+     * Checking if current device supports bluetooth.
```
shobhitagarwal1612: "Returns true if the device supports Bluetooth otherwise throws an IllegalStateException"

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothUtils.java
```diff
@@ -0,0 +1,76 @@
...
+     */
+    public static boolean isSupportBluetooth() {
```
shobhitagarwal1612: Method name should be `isBluetoothSupported`

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothUtils.java
```diff
@@ -0,0 +1,76 @@
...
+        if (getBluetoothAdapter() != null) {
+            return true;
+        } else {
+            throw new IllegalStateException("Your device does not support bluetooth!");
+        }
+    }
+
+    /**
+     * Checking if bluetooth is enabled.
```
shobhitagarwal1612: Similar to the above comment, please fix this method doc-string

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothUtils.java
```diff
@@ -0,0 +1,76 @@
...
+     */
+    public static boolean isBluetoothEnabled() {
+        return isSupportBluetooth() && getBluetoothAdapter().isEnabled();
+    }
+
+    /**
+     * Disable and enable the bluetooth.
```
shobhitagarwal1612: Shouldn't this be only "Disables bluetooth" since it does not enable it?

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothUtils.java
```diff
@@ -0,0 +1,76 @@
...
+    public static void enableBluetooth() {
+        if (!isBluetoothEnabled()) {
+            new Thread(() -> {
+                getBluetoothAdapter().enable();
+            }).start();
+        }
+    }
+
+    /**
+     * Checking if the current {@link android.app.Activity} is destroyed.
```
shobhitagarwal1612: "...is finished or about to be finished"

skunkworks_crow/src/main/java/org/odk/share/services/ReceiverService.java
```diff
@@ -55,12 +55,27 @@ private void addDownloadJobSubscription() {
 
     }
 
+    public void startDownloading(String macAddress) {
+        PersistableBundleCompat extras = new PersistableBundleCompat();
+        extras.putString("mac", macAddress);
+        extras.putBoolean("isBluetooth", true);
```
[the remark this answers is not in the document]
shobhitagarwal1612: @huangyz0918 Please use a common method to schedule jobs. In this case, rxEventBus is not triggered which happens in the other method(wifi hotspot).

skunkworks_crow/src/main/java/org/odk/share/tasks/DownloadJob.java
```diff
@@ -94,26 +103,60 @@ protected Result onRunJob(@NonNull Params params) {
...
+                Timber.d("Socket connected");
+                dis = new DataInputStream(new BufferedInputStream(socket.getInputStream()));
+                dos = new DataOutputStream(socket.getOutputStream());
+            }
+
+            rxEventBus.post(receiveForms());
+        } catch (IOException e) {
+            Timber.e(e);
+        }
+
```
shobhitagarwal1612: Please remove this empty line

skunkworks_crow/src/main/java/org/odk/share/utilities/PermissionUtils.java
```diff
@@ -0,0 +1,63 @@
...
+
+        builder.setNegativeButton(targetActivity.getString(R.string.cancel), (DialogInterface dialog, int which) -> {
+            dialog.dismiss();
+            targetActivity.finish();
+        });
+
+        builder.setCancelable(false);
+        builder.show();
+    }
+
```
shobhitagarwal1612: Please remove these blank lines

skunkworks_crow/src/main/java/org/odk/share/utilities/PermissionUtils.java
```diff
@@ -0,0 +1,63 @@
...
+    public static boolean isGPSEnabled(Activity targetActivity) {
+        LocationManager locationManager = (LocationManager) targetActivity.getSystemService(LOCATION_SERVICE);
+        boolean gpsEnabled = false;
+        if (locationManager != null) {
+            gpsEnabled = locationManager.isProviderEnabled(LocationManager.GPS_PROVIDER);
+        }
+
+        return gpsEnabled;
+    }
+
```
shobhitagarwal1612: Please only leave 1 line gap

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtReceiverActivity.java
```diff
@@ -0,0 +1,192 @@
...
+                }, Timber::e);
+    }
+
+    @Override
+    public void foundDevice(BluetoothDevice device) {
+        bluetoothListAdapter.addDevice(device);
+    }
+
+    @Override
+    public void onItemClick(BluetoothDevice dev) {
```
shobhitagarwal1612: Please use `device` instead of `dev` for better readability

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtSenderActivity.java
```diff
@@ -0,0 +1,150 @@
...
+        compositeDisposable.add(addBluetoothEventSubscription());
+    }
+
+    @Override
+    protected void onPause() {
+        super.onPause();
+        compositeDisposable.clear();
+    }
+}
+
```
shobhitagarwal1612: Please remove empty lines

skunkworks_crow/src/main/java/org/odk/share/views/ui/main/MainActivity.java
```diff
@@ -203,22 +224,18 @@ private boolean isCollectInstalled() {
...
-                try {
-                    startActivity(new Intent(Intent.ACTION_VIEW, Uri.parse("market://details?id=" + COLLECT_PACKAGE)));
-                } catch (android.content.ActivityNotFoundException e) {
-                    startActivity(new Intent(Intent.ACTION_VIEW, Uri.parse("https://play.google.com/store/apps/details?id=" + COLLECT_PACKAGE)));
-                }
+
+        builder.setPositiveButton(getString(R.string.install), (DialogInterface dialog, int which) -> {
+            try {
+                startActivity(new Intent(Intent.ACTION_VIEW, Uri.parse("market://details?id=" + COLLECT_PACKAGE)));
+            } catch (android.content.ActivityNotFoundException e) {
```
shobhitagarwal1612: Please import the class at the top of the class instead of using fully qualified name.

skunkworks_crow/src/main/java/org/odk/share/views/ui/send/fragment/BlankFormsFragment.java
```diff
@@ -156,13 +161,39 @@ private void setEmptyViewVisibility(int len) {
...
+        Intent bluetoothIntent = new Intent(getActivity(), BtSenderActivity.class);
+        setupSendingIntent(hotspotIntent);
```
shobhitagarwal1612: setup should happen after the TRANSFER_OPTION has been selected to prevent extra initialisation of the other intent.

skunkworks_crow/src/main/java/org/odk/share/views/ui/send/fragment/BlankFormsFragment.java
```diff
@@ -156,13 +161,39 @@ private void setEmptyViewVisibility(int len) {
...
+                                startActivity(bluetoothIntent);
+                                getActivity().finish();
+                            }
+                            break;
+                        case 1:
+                            if (getActivity() != null) {
+                                startActivity(hotspotIntent);
+                                getActivity().finish();
+                            }
+                            break;
```
shobhitagarwal1612: Switch cases have most of their code common. Please reuse the code.

skunkworks_crow/src/main/res/values/strings.xml
```diff
@@ -130,4 +130,19 @@
     <string name="odk_destination_dir_error">The destination should not be empty</string>
     <string name="location_permission_needed">Location Permission Needed</string>
     <string name="location_settings_dialog">Enable location from the settings.</string>
+
+    <!-- Bluetooth Related -->
```
shobhitagarwal1612: Comment not needed

skunkworks_crow/src/main/res/layout/list_item_bt_device.xml
```diff
@@ -0,0 +1,44 @@
...
+            android:id="@+id/tv_device_address"
+            style="@style/MatchParent"
+            android:layout_marginStart="6dp"
+            android:layout_marginLeft="6dp"
+            android:layout_marginTop="2dp"
+            android:layout_marginBottom="2dp"
+            android:gravity="center_vertical"
+            android:textSize="12sp"
+            tools:text="default device address" />
+
```
shobhitagarwal1612: Please remove blank line

skunkworks_crow/src/main/res/layout/list_item_bt_device.xml
```diff
@@ -0,0 +1,44 @@
...
+    </LinearLayout>
+
```
shobhitagarwal1612: same as above

skunkworks_crow/src/main/java/org/odk/share/views/ui/send/fragment/FilledFormsFragment.java
huangyz0918: Fixed.

skunkworks_crow/src/main/java/org/odk/share/views/ui/send/fragment/FilledFormsFragment.java
huangyz0918: Fixed.

skunkworks_crow/src/main/java/org/odk/share/views/ui/send/fragment/BlankFormsFragment.java
huangyz0918: Fixed.

skunkworks_crow/src/main/java/org/odk/share/views/ui/main/MainActivity.java
huangyz0918: Fixed.

skunkworks_crow/src/main/java/org/odk/share/views/ui/main/MainActivity.java
huangyz0918: Fixed.

skunkworks_crow/src/main/java/org/odk/share/views/ui/main/MainActivity.java
huangyz0918: Fixed.

skunkworks_crow/src/main/java/org/odk/share/views/ui/instance/fragment/ReviewedInstancesFragment.java
huangyz0918: Sorry for forgetting that, thanks for reminding. 
Fixed.

skunkworks_crow/src/main/java/org/odk/share/views/ui/instance/InstancesList.java
huangyz0918: Fixed.

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothReceiver.java
huangyz0918: Fixed.

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothReceiver.java
huangyz0918: Removed.

huangyz0918: But I think some of them can be used for future UI updating.

skunkworks_crow/src/main/java/org/odk/share/tasks/DownloadJob.java
huangyz0918: So you mean we should call `cancel()` to cancel the job here? @lakshyagupta21

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothUtils.java
huangyz0918: Improved.

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothUtils.java
huangyz0918: hmmm... maybe not, I'm not sure. 
I deleted that in a new update and it works the same as before.

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothUtils.java
huangyz0918: deleted.

skunkworks_crow/src/main/java/org/odk/share/tasks/DownloadJob.java
huangyz0918: Fixed.

skunkworks_crow/src/main/res/values/strings.xml
huangyz0918: Fixed.

skunkworks_crow/src/main/res/layout/list_item_bt_device.xml
huangyz0918: Fixed.

skunkworks_crow/src/main/res/layout/list_item_bt_device.xml
huangyz0918: Fixed.

skunkworks_crow/src/main/java/org/odk/share/views/ui/send/fragment/BlankFormsFragment.java
huangyz0918: Fixed.

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothUtils.java
huangyz0918: Renamed.

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothUtils.java
huangyz0918: Fixed.

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothUtils.java
huangyz0918: Improved.

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothReceiver.java
huangyz0918: Refactored.

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothReceiver.java
huangyz0918: Done.

skunkworks_crow/src/main/AndroidManifest.xml
huangyz0918: Fixed.

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothUtils.java
huangyz0918: Fixed.

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothUtils.java
huangyz0918: Fixed.

skunkworks_crow/src/main/java/org/odk/share/tasks/DownloadJob.java
huangyz0918: Removed.

skunkworks_crow/src/main/java/org/odk/share/tasks/DownloadJob.java
huangyz0918: Okey, see updated.

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BluetoothListAdapter.java
huangyz0918: See updated.

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtReceiverActivity.java
huangyz0918: Fixed.

skunkworks_crow/src/main/java/org/odk/share/utilities/PermissionUtils.java
huangyz0918: removed.

skunkworks_crow/src/main/java/org/odk/share/utilities/PermissionUtils.java
huangyz0918: removed.

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtReceiverActivity.java
huangyz0918: improved.

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtSenderActivity.java
huangyz0918: removed.

skunkworks_crow/src/main/java/org/odk/share/views/ui/main/MainActivity.java
huangyz0918: You mean, static import? Done.

huangyz0918:
> Also one thing that I noticed while scanning the devices is that it is showing all the devices which was paired before to the device but not available at the moment of scanning, so please take a look at that as well.

Yeah, I want to show the paired devices, that can provide a quick way for our user to connect. It doesn't matter if the device is unreachable, if the device is reachable, the transmission will start if the connection is established. I updated the UI in a new way, thing will make more sense now.

skunkworks_crow/src/main/java/org/odk/share/tasks/UploadJob.java
huangyz0918: Yes, thanks for reminding. Fixed.

skunkworks_crow/src/main/java/org/odk/share/services/SenderService.java
huangyz0918: fixed.

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothReceiver.java
```diff
@@ -0,0 +1,60 @@
...
+    @Override
+    public void onReceive(Context context, Intent intent) {
+        String action = intent.getAction();
+        if (action == null) {
+            return;
+        }
+
+        Timber.d(TAG, action);
+        BluetoothDevice dev = intent.getParcelableExtra(BluetoothDevice.EXTRA_DEVICE);
+        if (dev != null) {
```
lakshyagupta21: Give some meaningful name like `bluetoothDevice` or just a `device`

lakshyagupta21: And what if dev is null, on `deviceFound `should not be called.

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothReceiver.java
```diff
@@ -49,47 +36,18 @@ public void onReceive(Context context, Intent intent) {
         Timber.d(TAG, action);
         BluetoothDevice dev = intent.getParcelableExtra(BluetoothDevice.EXTRA_DEVICE);
         if (dev != null) {
-            Timber.d(TAG + "BluetoothDevice: " + dev.getName() + ", " + dev.getAddress());
+            Timber.d("%s BluetoothDevice: %s, Address: %s", TAG, dev.getName(), dev.getAddress());
```
shobhitagarwal1612: TAG is redundant with Timber as it is already included. So, please remove it

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothReceiver.java
```diff
@@ -49,47 +36,18 @@ public void onReceive(Context context, Intent intent) {
...
         }
         switch (action) {
             case BluetoothAdapter.ACTION_STATE_CHANGED:
                 int state = intent.getIntExtra(BluetoothAdapter.EXTRA_STATE, 0);
-                Timber.d(TAG + "STATE: " + state);
-                break;
-            case BluetoothAdapter.ACTION_DISCOVERY_STARTED:
+                Timber.d("%s STATE: %s", TAG, state);
```
shobhitagarwal1612: Same as above

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothUtils.java
lakshyagupta21: I think @shobhitagarwal1612 wants to say if this method is used to check whether bluetooth is supported or not, so it is expected to return either tue or false. It shouldn't throw any Exception and if it throws then the caller should be catching or handling it.

skunkworks_crow/src/main/java/org/odk/share/tasks/DownloadJob.java
```diff
@@ -94,26 +104,61 @@ protected Result onRunJob(@NonNull Params params) {
...
+            case Share.TransferMethod.HOTSPOT:
+                ip = params.getExtras().getString(IP, "");
+                port = params.getExtras().getInt(PORT, -1);
+                break;
+            case Share.TransferMethod.BLUETOOTH:
+                targetMacAddress = params.getExtras().getString("mac", null);
+                break;
+        }
+
+        setupDataStreamsAndReceive(method, targetMacAddress);
```
lakshyagupta21: If you kept `targetMacAddress `as global then there is no need to pass.

skunkworks_crow/src/main/java/org/odk/share/tasks/DownloadJob.java
lakshyagupta21: I don't see any checks for `bluetoothDevice `and `bluetoothAdapter `null checks

skunkworks_crow/src/main/java/org/odk/share/tasks/DownloadJob.java
lakshyagupta21: No, I'm trying to say if it's not able to connect to the socket then it will through the IOException, in that case, we should stop the form transfer and for that, we need to cancel/stop the scheduled job.

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothReceiver.java
```diff
@@ -0,0 +1,60 @@
...
+                Timber.d("EXTRA_RSSI: %s", rssi);
+                bluetoothReceiverListener.onDeviceFound(dev);
+                break;
+        }
+    }
+
+    /**
+     * Listener for bluetooth devices when we found a new {@link BluetoothDevice}.
+     */
+    public interface BluetoothReceiverListener {
```
shobhitagarwal1612: Why aren't we broadcasting the discovery status as well, something like 'onDiscoveryStarted()' and 'onDiscoveryStopped()'? That might work as a trigger for initiating bluetooth related actions instead of assuming that enable bluetooth works without any fault every time.

skunkworks_crow/src/main/java/org/odk/share/tasks/DownloadJob.java
```diff
@@ -146,18 +188,28 @@ private DownloadEvent receiveForms() {
         return new DownloadEvent(DownloadEvent.Status.FINISHED, sbResult.toString());
     }
 
+    /**
+     * Close all the connections.
+     */
+    private void closeConnections() throws IOException {
+        if (socket != null) {
```
lakshyagupta21: No sure if closing socket before their dis and dos will hve any affect or not, but to be safe we can  close it after closing dis and dos.

skunkworks_crow/src/main/java/org/odk/share/tasks/UploadJob.java
```diff
@@ -89,36 +97,75 @@ protected Result onRunJob(@NonNull Params params) {
...
-            Timber.d("Start Sending");
-            processSelectedFiles(instancesToSend);
+            rxEventBus.post(uploadInstances());
+        } catch (IOException e) {
+            Timber.e(e);
+        }
+    }
 
-            // close connection
+    private void closeConnections() throws IOException {
```
lakshyagupta21: You forgot to close the `serverSocket`.

skunkworks_crow/src/main/java/org/odk/share/services/ReceiverService.java
```diff
@@ -52,15 +52,29 @@ private void addDownloadJobSubscription() {
                             break;
                     }
                 }).subscribe();
+    }
 
+    public void startBtDownloading(String macAddress) {
+        PersistableBundleCompat extras = new PersistableBundleCompat();
+        extras.putString("mac", macAddress);
+        extras.putInt("MODE_OF_TRANSFER", Share.TransferMethod.BLUETOOTH);
+        JobRequest request = new JobRequest.Builder(DownloadJob.TAG)
```
shobhitagarwal1612: Why not pass the `extras` to `startJob()` method? Also, like I mentioned earlier that would resolve the problem that this job doesn't trigger a rx bus event

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BluetoothListAdapter.java
```diff
@@ -0,0 +1,142 @@
...
+    @Override
+    public ViewHolder onCreateViewHolder(@NonNull ViewGroup parent, int viewType) {
+        View view = LayoutInflater.from(parent.getContext()).inflate(R.layout.list_item_bt_device, null);
+        return new ViewHolder(view);
+    }
+
+    /**
+     * Rescan the bluetooth devices and update the list.
+     */
+    public void updateDeviceList() {
```
lakshyagupta21: Lets not keep this method as well in adapter, try to keep it in some utility class or the activity it is used in.

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtReceiverActivity.java
```diff
@@ -0,0 +1,193 @@
...
+        ButterKnife.bind(this);
+        initEvents();
+
+        setTitle(getString(R.string.connect_bluetooth_title));
+        setSupportActionBar(toolbar);
+
+        // checking for if bluetooth enabled
+        if (!BluetoothUtils.isBluetoothEnabled()) {
+            BluetoothUtils.enableBluetooth();
+            bluetoothListAdapter.updateDeviceList();
```
lakshyagupta21: The `updateDevieList()` should be present here in this activity.

skunkworks_crow/src/main/java/org/odk/share/views/ui/instance/InstancesList.java
```diff
@@ -119,13 +124,29 @@ private void onListItemClick(View view, int position) {
...
+                    switch (which) {
+                        case 0:
+                            intent.setClass(this, BtSenderActivity.class);
+                            break;
+                        case 1:
+                            intent.setClass(this, HpSenderActivity.class);
+                            break;
+                    }
+                    startActivity(intent);
+                    finish();
```
lakshyagupta21: Why calling finish() ?

skunkworks_crow/src/main/java/org/odk/share/views/ui/main/MainActivity.java
```diff
@@ -103,10 +106,26 @@ private void setupAdapter() {
...
+                    switch (which) {
+                        case 0:
+                            intent.setClass(this, BtReceiverActivity.class);
+                            break;
+                        case 1:
+                            intent.setClass(this, HpReceiverActivity.class);
+                            break;
+                    }
+                    startActivity(intent);
+                    finish();
```
lakshyagupta21: Same as above why calling finish()?

skunkworks_crow/src/main/java/org/odk/share/views/ui/send/fragment/BlankFormsFragment.java
```diff
@@ -156,13 +160,34 @@ private void setEmptyViewVisibility(int len) {
...
+                            case 0:
+                                intent.setClass(getActivity(), BtSenderActivity.class);
+                                break;
+                            case 1:
+                                intent.setClass(getActivity(), HpSenderActivity.class);
+                                break;
+                        }
+                        setupSendingIntent(intent);
+                        startActivity(intent);
+                        getActivity().finish();
```
lakshyagupta21: Same as above

skunkworks_crow/src/main/java/org/odk/share/views/ui/send/fragment/FilledFormsFragment.java
```diff
@@ -150,13 +154,34 @@ private void setEmptyViewVisibility(int len) {
...
+                            case 0:
+                                intent.setClass(getActivity(), BtSenderActivity.class);
+                                break;
+                            case 1:
+                                intent.setClass(getActivity(), HpSenderActivity.class);
+                                break;
+                        }
+                        setupSendingIntent(intent);
+                        startActivity(intent);
+                        getActivity().finish();
```
lakshyagupta21: Same as above.

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothReceiver.java
huangyz0918: fixed.

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothReceiver.java
huangyz0918: fixed.

skunkworks_crow/src/main/java/org/odk/share/tasks/DownloadJob.java
huangyz0918: sure.

skunkworks_crow/src/main/java/org/odk/share/tasks/DownloadJob.java
huangyz0918: updated.

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothReceiver.java
huangyz0918: Resolved.

huangyz0918: Fixed.

skunkworks_crow/src/main/java/org/odk/share/tasks/DownloadJob.java
huangyz0918: Fixed.

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BluetoothListAdapter.java
huangyz0918: sure

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtReceiverActivity.java
huangyz0918: fixed.

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothReceiver.java
huangyz0918: Sure, we can update the ui as well.

skunkworks_crow/src/main/java/org/odk/share/services/ReceiverService.java
huangyz0918: improved.

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothUtils.java
lakshyagupta21: @huangyz0918 Either add throws in method signature and let its caller handle it or don't throw any exception just return `false`

skunkworks_crow/src/main/java/org/odk/share/tasks/UploadJob.java
```diff
@@ -89,36 +97,78 @@ protected Result onRunJob(@NonNull Params params) {
...
+            bluetoothServerSocket.close();
+        }
+    }
+
+    private UploadEvent uploadInstances() {
+        try {
+            // show dialog and connected
+            Timber.d("Start Sending");
+            processSelectedFiles(instancesToSend);
+            closeConnections();
```
lakshyagupta21: Add `closeConnections()` in finally block.

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtReceiverActivity.java
```diff
@@ -0,0 +1,228 @@
...
+        bluetoothAdapter = BluetoothAdapter.getDefaultAdapter();
+        LinearLayoutManager linearLayoutManager = new LinearLayoutManager(this);
+        DividerItemDecoration dividerItemDecoration = new DividerItemDecoration(recyclerView.getContext(),
+                linearLayoutManager.getOrientation());
+        recyclerView.addItemDecoration(dividerItemDecoration);
+        recyclerView.setLayoutManager(linearLayoutManager);
+        recyclerView.setAdapter(bluetoothListAdapter);
+        bluetoothReceiver = new BluetoothReceiver(this, this);
+        addPairedDevices();
+        BluetoothAdapter.getDefaultAdapter().startDiscovery();
```
lakshyagupta21: Either use `BluetoothAdapter.getDefaultAdapter()` everywhere or if you're assigning it to `bluetoothAdapter` then use the same variable everywhere it needs to be consistent everywhere.

huangyz0918:
@lakshyagupta21 @shobhitagarwal1612 I have improved the code as your suggested, please have a new round review and make sure the code is good to merge, thanks. 

I have tested the new feature using sender/receiver with success:
1. Google Nexus 6P Android 8.0.0 (sender/receiver) 
2. Google Nexus 6P Android 8.1.0 (sender/receiver)
3. Google Nexus 5 Android 6.0.1 (sender/receiver)
4. Meizu M5 Android 6.0 (sender/receiver)

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothReceiver.java
```diff
@@ -33,21 +34,28 @@ public void onReceive(Context context, Intent intent) {
...
-                Timber.d("EXTRA_RSSI: %s", rssi);
-                bluetoothReceiverListener.onDeviceFound(dev);
-                break;
+        BluetoothDevice bluetoothDevice = intent.getParcelableExtra(BluetoothDevice.EXTRA_DEVICE);
+        if (bluetoothDevice != null) {
+            Timber.d("BluetoothDevice: %s, Address: %s", bluetoothDevice.getName(), bluetoothDevice.getAddress());
+            switch (action) {
+                case BluetoothAdapter.ACTION_STATE_CHANGED:
+                    int state = intent.getIntExtra(BluetoothAdapter.EXTRA_STATE, 0);
+                    Timber.d("STATE: %s", state);
```
shobhitagarwal1612: What is the purpose of this? I guess it is also not being used.

skunkworks_crow/src/main/java/org/odk/share/tasks/DownloadJob.java
```diff
@@ -121,27 +121,29 @@ private void initJob(Params params) {
...
-                    bluetoothSocket.connect();
-                }
+                    if (!bluetoothSocket.isConnected()) {
+                        bluetoothSocket.connect();
+                    }
 
-                if (bluetoothSocket.isConnected()) {
-                    rxEventBus.post(new BluetoothEvent(BluetoothEvent.Status.CONNECTED));
-                }
+                    if (bluetoothSocket.isConnected()) {
```
shobhitagarwal1612: this can be replaced with `else`

lakshyagupta21: I don't this should go in else because previous if block says it to connect so I think this `if` stmt check can be removed and we can directly write 
` rxEventBus.post(new BluetoothEvent(BluetoothEvent.Status.CONNECTED));`
as it is expected that
```
if (!bluetoothSocket.isConnected()) {
    bluetoothSocket.connect();
}
```
will ensure that bluetooth is connected and if by any chance it fails then it should throw some error or callback that we need to handle explicitly.

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothReceiver.java
huangyz0918: removed.

skunkworks_crow/src/main/java/org/odk/share/tasks/DownloadJob.java
huangyz0918: agreed with @lakshyagupta21 , just removing the `if-else`.

shobhitagarwal1612: Is `bluetoothSocket.connect();` a synchronous event? Are there any callbacks for bluetooth connected or disconnected? If not then we need to fix it.

shobhitagarwal1612: Second problem is that we shouldn't establish bluetooth socket connection within `DownloadJob`. DownloadJob's purpose is to start download task and it shouldn't be doing anything else. This would testing the code in the later part difficult.

huangyz0918: > Is `bluetoothSocket.connect();` a synchronous event? Are there any callbacks for bluetooth connected or disconnected? If not then we need to fix it.

This method is synchronous yes, and if the connection cannot be established, it will throw an IO Exception. So we don't have to fix that.

huangyz0918: > Second problem is that we shouldn't establish bluetooth socket connection within `DownloadJob`. DownloadJob's purpose is to start download task and it shouldn't be doing anything else. This would testing the code in the later part difficult.

An issue is, we cannot pass an object as a parameter into our jobs, I tried to establish the socket connection in the activity but it seems the job is not relay on that. Besides, the socket and stream for hotspot are created in the job as well, so I think it doesn't matter.

lakshyagupta21: @shobhitagarwal1612 Can you suggest a forward, I tried the workarounds but I'm not able to think one because DownloadJob needs socket objects and we cannot pass the socket objects directly in the `DownloadJob` and `UploadJob`.

shobhitagarwal1612: Okay. I'll give it a try and post it here.

huangyz0918:
I notice an issue when we are trying to send form instances, using two Google Nexus 6P (android 8.0.0 and android 8.0.1): The progress will stuck when we send form instance(s), I cannot use hotspot method to verify where this issue comes from (since wifi cannot be established in android 8.0.1, the hotspot cannot be turned on and the QR Code didn't show up either), so can you help me with that issue? @lakshyagupta21 @shobhitagarwal1612 I don't know if you have notice the android 8+ issue with hotspot, and I want to validate if the issue comes from here and fix that.

lakshyagupta21:
@huangyz0918 I'm still not sure about this PR whether it will break anything or not, and I need some more time to test it and I know that some of your tasks are dependent on this PR and @shobhitagarwal1612 is also thinking about solving the above issue. So we can merge this PR in some intermediate branch and then you can work considering that as your development branch. This way you'll get unblocked.
What do you say @shobhitagarwal1612 and @huangyz0918 ?

huangyz0918:
> So we can merge this PR in some intermediate branch and then you can work considering that as your development branch.

Good idea to merge that into a `dev` branch. @lakshyagupta21 and I think you can have tests in the `dev` until you think it's ready for merging into `master`. How about @shobhitagarwal1612 , what do you think?

shobhitagarwal1612:
I think merging into a `dev` branch is not a good idea as this PR touches a lot of code that would create an extra overhead of keeping up that branch with the latest master branch. So, all of the subsequent PRs will still have lots of conflicts. So lets make sure that this PR does not disrupt the existing wifi-hotspot transfer functionality and merge this as it is. Then we can test this everyday and open issues as we find them.

skunkworks_crow/src/test/java/org/odk/share/activities/BtReceiverActivityTest.java
```diff
@@ -0,0 +1,107 @@
...
+     */
+    @Test
+    public void testBroadcastReceiverRegistered() {
+        List<ShadowApplication.Wrapper> registeredReceivers = ShadowApplication.getInstance().getRegisteredReceivers();
+
+        assertFalse(registeredReceivers.isEmpty());
+
+        boolean receiverFound = false;
+        for (ShadowApplication.Wrapper wrapper : registeredReceivers) {
+            if (!receiverFound) {
```
lakshyagupta21: Its always going to be false first so whenever it gets inside the for each loop then the receiver will be assigned, then there is no use of iterating over registered receivers.\
Anyways, the code requires clean up that we can do later.

lakshyagupta21:
@huangyz0918 @shobhitagarwal1612 Merging this PR, I did some testing on Bluetooth form transfer from Android 7 to Android 9, but I was not able to transfer the other way. And there were some issues related to UI/UX as well I think it is not in parity with the Hotspot functionality, like
1. before pressing back while form transfer it should show some dialog to get the confirmation for stopping the connection.
2. BlueTooth is still 'ON' even if the connection is complete/canceled.
3. I'm not able to connect to new bluetooth directly via an app and it doesn't show me any new device.
These are the basic issues which I think user will face, there can be more.

huangyz0918:
> I'm not able to connect to new bluetooth directly via an app and it doesn't show me any new device.

I think you should wait longer since the bluetooth classic takes a very long time to discovery some other devices. I'm not sure how to improve that, any suggestions? @lakshyagupta21

huangyz0918:
> BlueTooth is still 'ON' even if the connection is complete/canceled.
I can fix this, and I'll open an issue for that.

PR #261, opened 2019-06-08
set minimum password length to enable OK button.  First fix version.

jhbiggs:
Hi, I've answered the prompts below.  This is my first real pull request.  Thanks for your patience in advance.

Closes #

I have run the application on an emulator using API 24.  I didn't update the gradle version as it would have caused changes to those files, and potential extraneous changes to merge.

I looked at the TextWatcher as an automatic class for observing any changes.  I sought any possible classes which would work here, this seemed like the best option aside from introducing global dialog changes.

When the AlertDialog launches, users will be required to enter at least 8 characters to enable the OK button.  Some potential risks might include: the user doesn't know this requirement and doesn't follow through enough to get to the 8 character switch.  The user may escape out of the alertDialog without changing the password.

### DIFF
--- a/skunkworks_crow/src/main/java/org/odk/share/views/ui/settings/SettingsActivity.java
+++ b/skunkworks_crow/src/main/java/org/odk/share/views/ui/settings/SettingsActivity.java
@@ -1,5 +1,6 @@
 package org.odk.share.views.ui.settings;
 
+import android.content.DialogInterface;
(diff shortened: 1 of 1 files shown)


jhbiggs:
Right, so the codeCheck set minimum gradle build over the current repository.  I think that's why my build_debug CircleCI failed.  I'm updating the gradle build and will try again.

PR #272, opened 2019-07-21
[MRG] Improved the bluetooth scanning progress

### LINKED ISSUE #271: Improve the refresh progress in bluetooth device page
#### Problem description
In the bluetooth pairing page, we can see a lot of devices (paired and unpaired), but the scanning progress may be very slow with bluetooth classic, I think we should find an approach to improve that. On the other hand, we don't have a scanning timeout or any UI changes when scan progress starting... So I think there are still some work left in this page. 

#### Expected behavior
- Speed up the paring progress and scanning progress if possible.
- Improve the UI when discovering other devices.

#### Other information 
@opendatakit-bot claim

huangyz0918:
Closes #271 

I tested using Nexus 6P (android 8.0.1) everything went well.

The screenshots looks like:

|scanning page|sender page|
|:--:|:--:|
|||

And I built a fixed time that bluetooth device is discoverable, in this way, we can know when we has finished the scanning and get the completed scanned results. An issue is should we notify the users that the discoverable time is 60s?

N/A
N/A


skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothReceiver.java
```diff
@@ -33,25 +33,27 @@ public void onReceive(Context context, Intent intent) {
...
-                case BluetoothAdapter.ACTION_DISCOVERY_STARTED:
-                    Timber.d("bluetooth devices discovery started...");
-                    bluetoothReceiverListener.onDiscoveryStarted();
-                    break;
-                case BluetoothAdapter.ACTION_DISCOVERY_FINISHED:
-                    Timber.d("bluetooth devices discovery finished...");
-                    bluetoothReceiverListener.onDiscoveryFinished();
-                    break;
+        // if the discovery started, update the ui in activity.
+        if (BluetoothAdapter.ACTION_DISCOVERY_STARTED.equals(action)) {
```
lakshyagupta21: I think `switch `case would be better in this case, it will make it easy to understand.

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtSenderActivity.java
```diff
@@ -133,6 +136,16 @@ private Disposable addUploadEventSubscription() {
                 }, Timber::e);
     }
 
+    /**
+     * Enable the bluetooth discovery for other devices. The timeout is 60 seconds.
+     */
+    private void enableDiscovery() {
```
lakshyagupta21: And I think sender should be made `DISCOVERABLE` not the receiver so that it would be able to see the sender in its scanned results.

skunkworks_crow/src/main/res/values/strings.xml
```diff
@@ -139,8 +139,10 @@
     <string name="connect_bluetooth_title">Device to Receive</string>
     <string name="connected_bluetooth_downloading">Connected, start downloading&#8230;</string>
     <string name="connecting_title">Connecting</string>
-    <string name="bluetooth_paired">paired, may unreachable</string>
-    <string name="bluetooth_unpaired">unpaired, reachable</string>
+    <string name="scanning_title">Scanning</string>
+    <string name="scanning_msg">Scanning bluetooth devices, please wait&#8230;</string>
+    <string name="bluetooth_paired">paired</string>
```
lakshyagupta21: It could be like `Device Paired` or `Device Unpaired`.

huangyz0918:
@lakshyagupta21 @shobhitagarwal1612 please review, thanks.

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothReceiver.java
huangyz0918: I have an issue using `swith-case`, the `ACTION_DISCOVERY_STARTED` cannot be received, so I changed them to `if-else`, I think that is not a big deal.

huangyz0918: That's odd, maybe related to the `break`.

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtSenderActivity.java
huangyz0918: This method allows sender to be `DISCOVERABLE` for a specific duration, by default, it is 120s. And I set a fixed 60s for that.  I think this change can help receiver to show all the possible senders and help them establish a valid connection in a fixed time, since in the next step we are going  to build a default method, we need that timeout I think. 

As for who should scan bluetooth devices, I think sender should host a file server so it should be `DISCOVERABLE`, the connect invitation should be sent by the receiver.

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothReceiver.java
lakshyagupta21: That's strange, okay I'll take a look, anyways if you've added `if` statements then it would be better if you can make these individual stmts as `if else if..`, right now they all are written as individual `if` statements.

skunkworks_crow/src/main/res/values/strings.xml
huangyz0918: Sure, will fix right now.

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothReceiver.java
huangyz0918: Thanks, no hurry, let me have a try with `switch`.

skunkworks_crow/src/main/java/org/odk/share/bluetooth/BluetoothReceiver.java
```diff
@@ -33,24 +33,26 @@ public void onReceive(Context context, Intent intent) {
...
 
-        BluetoothDevice bluetoothDevice = intent.getParcelableExtra(BluetoothDevice.EXTRA_DEVICE);
-        if (bluetoothDevice != null) {
-            Timber.d("BluetoothDevice: %s, Address: %s", bluetoothDevice.getName(), bluetoothDevice.getAddress());
-            switch (action) {
-                case BluetoothDevice.ACTION_FOUND:
+        switch (action) {
+            case BluetoothAdapter.ACTION_DISCOVERY_STARTED:
+                // if the discovery started, update the ui in activity.
+                Timber.d("bluetooth devices discovery started...");
```
huangyz0918: Hmm.. It seems `ACTION_DISCOVERY_STARTED` should write before `ACTION_FOUND`, I'm not sure, but it works now @lakshyagupta21 .

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtSenderActivity.java
lakshyagupta21: I don't think just by hosting the file the sender is made discoverable.  Please check that on multiple devices.

huangyz0918: Bluetooth doesn't support multiple devices, the files server should be ready first, so the sender should get ready before the receiver. The connection order should start with the sender then the receiver, so letting receiver to connect a ready server is no issues.

lakshyagupta21: By multiple devices, I mean to say that try to test this behavior in different devices, not by running them all at once.

huangyz0918:
New updates:

- Added a count down timer for discoverable time.
- Added a cancel button while connecting to bluetooth device.

@lakshyagupta21 @shobhitagarwal1612 This is an important PR, and I think it will be good if you can merge ASAP, thanks.

PR #274, opened 2019-07-22
[MRG] Added Unit Tests for Preference, etc

### LINKED ISSUE #265: Write the unit tests for the bluetooth feature.
### Problem description 
For bluetooth feature added in #260 , we need to write the unit tests about the bluetooth devices pairing and connecting, even some future UI works.

#### Expected behavior
Build the unit tests and add them into the Gradle tasks for automatic building and CI testing.

#### Other information 
@opendatakit-bot claim

huangyz0918:
Closes #265 

I have tested the unite test locally, everything went well.

N/A
N/A


skunkworks_crow/src/test/java/org/odk/share/activities/HpReceiverActivityTest.java
```diff
@@ -0,0 +1,52 @@
...
+import org.robolectric.annotation.LooperMode;
+import org.robolectric.shadows.ShadowLog;
+
+import static junit.framework.TestCase.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.robolectric.annotation.LooperMode.Mode.PAUSED;
+
+@RunWith(RobolectricTestRunner.class)
+@LooperMode(PAUSED)
+public class HpReceiverActivityTest {
```
lakshyagupta21: It would be awesome if we can add some cases to verify the form transfer, not necessary in this PR, we can do that in different but we should do that as well, we may have to take help of Mockito in that case to mock the sender's behaviour.

PR #276, opened 2019-07-23
[MRG] Removed the unreachable bluetooth devices.

### LINKED ISSUE #275: Bluetooth scanning results should not contain unreachable devices
#### Problem description
In the bluetooth devices page, we have all the paired devices shown, but some of them are unreachable so we need to remove them. 

Open the skunkworks-crow, click `receive` using bluetooth, we can see some scanned devices and some of them are just bonded but unreachable. 

#### Expected behavior
We should show all available devices.

#### Other information 
@opendatakit-bot claim

huangyz0918:
Closes #275 

I just removed the `addBondedDevices()` method, since we don't need all the paired devices. 
Those changes tested well in my Nexus 6P and Samsung Galaxy note3. 

Though we removed the `addBondedDevices()` method, but we can still see the differences between the paired devices and the unpaired devices. So I think this is the best solution.  

N/A

### DIFF
--- a/skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BluetoothListAdapter.java
+++ b/skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BluetoothListAdapter.java
@@ -21,6 +21,7 @@ import butterknife.BindView;
 import butterknife.ButterKnife;
 
 import static android.bluetooth.BluetoothDevice.BOND_BONDED;
+import static android.bluetooth.BluetoothDevice.BOND_NONE;
 
 
 /**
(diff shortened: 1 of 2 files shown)


huangyz0918:
This quick fix is ready for a review @lakshyagupta21 .

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BluetoothListAdapter.java
```diff
@@ -69,7 +70,7 @@ public void onBindViewHolder(@NonNull final ViewHolder viewHolder, int position)
         int deviceBondState = device.getBondState();
         viewHolder.deviceName.setText(deviceName == null ? viewHolder.itemView.getResources().getString(R.string.bluetooth_instance_name_default) : deviceName);
         viewHolder.deviceAddress.setText(String.format("%s (%s)", deviceAddress,
-                deviceBondState == 10 ? context.getString(R.string.bluetooth_unpaired) : context.getString(R.string.bluetooth_paired)));
```
lakshyagupta21: Just a question what was this hardcoded 10 value?

huangyz0918: The same as `BOND_NONE`.

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BluetoothListAdapter.java
```diff
@@ -69,7 +70,7 @@ public void onBindViewHolder(@NonNull final ViewHolder viewHolder, int position)
...
+                deviceBondState == BOND_NONE ? context.getString(R.string.bluetooth_unpaired) : context.getString(R.string.bluetooth_paired)));
         if (device.getBondState() == BOND_BONDED) {
```
lakshyagupta21: use `deviceBondState` in place of `device.getBondState()`

PR #270, opened 2019-07-18
[MRG] dialogs to the bluetooth sender/receiver

huangyz0918:
Closes #268 
Closes #269 
Closes #297 

#### What has been done to verify that this works as intended?
built an alert dialog for asking if we should disable the bluetooth. And I have tested in my Nexus 6P (Android 8.0.1) and Samsung Galaxy Note3 (Android 5.0), everything goes well.

attached please find screenshots when press `back` button on the devices.

|||
|:--:|:--:|
|[](https://i.loli.net/2019/07/18/5d2fecea395bc15491.png)|[](https://i.loli.net/2019/07/18/5d2fecea4dad177662.png)|

#### Why is this the best possible solution? Were any other approaches considered?
I added two dialog when called `onBackPressed`, and this is a best way to handle that issue.

#### How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?
N/A

#### Before submitting this PR, please make sure you have:
- [x] run `./gradlew checkCode` and confirmed all checks still pass OR confirm CircleCI build passes
- [x] verified that any code or assets from external sources are properly credited in comments and/or in the [about file](https://github.com/opendatakit/share/blob/master/share_app/src/main/assets/open_source_licenses.html).

PR #281, opened 2019-07-26
[MRG] Fix the result presenting

### LINKED ISSUE #280: Dialog shows null when sending all the existed forms
#### Problem description
When we are trying to send the existed forms (all) using WiFi hotspot, we can see a blank `transfer result` dialog. 

- click send/receive using wifi hotspot.
- choose all the existed forms (for receiver) in the sender.
- connect to each other and start transferring.
- see the blank alert dialog.

#### Expected behavior
the alert dialog should not show blank when sending all the existed forms. It should show something like "form XXX is exists in the receiver, transmission canceled."

#### Other information 
@opendatakit-bot claim

### LINKED ISSUE #293: Improve the result showing using bluetooth
#### Problem description
According to the [comment](https://github.com/opendatakit/skunkworks-crow/pull/281#issuecomment-518481636), we know once we use bluetooth as our transfer method, we will show both toasts and print the results in the activity, it's a bad UI work and we should keep that the same as hotspot, build a result dialog for that. 

- Send/Receive the forms using bluetooth as the main transfer method.
- See the results 

#### Expected behavior
- [ ] For bluetooth receivers, after receiving data, we should show the result dialog and finish the activity if user click `OK` or so.
- [ ] For the bluetooth sender, we should not present the sending result in the activity, we should also show the result dialog and finish the activity if needed.

#### Other information 
@opendatakit-bot claim

huangyz0918:
Closes #280 
Closes #293 

Tested with success in Nexus 6P (android 8.0.1).

Updated with this pic:

Will finish the activities after click `ok`.

This is the fastest and easiest way to fix this issue. We have another case: some forms are not existed and others are duplicated, in this case we only show the transferred forms, only when all the forms are duplicated, we show the message "form already exist \n please send different form".

N/A


lakshyagupta21:
@huangyz0918 I'm not able to send any forms from Android 9 to Android 7.1 and vice versa. And I don't have any other device to test on so let's fix the Android 9 issue first then maybe we can move to this PR for testing. Or if @shobhitagarwal1612 can help us here in testing it, then it would be awesome.

Also, I'm not able to see Android 9 device (sender) in receiver's list, it could be an issue related to making it `discoverable`.

huangyz0918:
Neither can I to discover the sender, but I think this issue was fixed in PR #272 , so don't worry about that. I have device in android 8.0.0, 8.0.1 and 5.0, they all work well. I'll flash my Nexus 6P to android 9 after current PRs tested and merged @lakshyagupta21 .

skunkworks_crow/src/main/res/values/strings.xml
```diff
@@ -148,7 +148,7 @@
     <string name="send_instance_title">Send Data</string>
     <string name="bluetooth_instance_name_default">Bluetooth Device</string>
     <string name="tv_sender_wait_for_connect">waiting for connection&#8230; \n will start transfer when connected </string>
-    <string name="tv_form_already_exist">form already exists \n please send different form</string>
+    <string name="tv_form_already_exist">form already exist \n please send different form</string>
```
lakshyagupta21: A message should not be printed like asking the user to send a different form, I think `Form already exists at receiver` will give the user enough information about any form transaction.

And this message should be appended with each and every form we send. Lets say if we send 3 forms with name form1, form2, form3 and all forms exists there, then we should be printing
form1- Form already exists at receiver's
form2- Form already exists at receiver's
form3- Form already exists at receiver's
We should give this information for each of the individual form transaction.

huangyz0918:
I think it's ready for a review @lakshyagupta21 .

lakshyagupta21:
When I try to send the blank form from Android 9 to Android 7.1 (previously I transferred the same form from Android 7.1 to Android 9), after form transfer it directly goes back to `Device to receive` screen without showing the status.

The status is shown to the user in the form of Toast which also pops up 2-3 times, which is `Download Queued` and `Form sent with success`. This toast is shown in the receiver's device. I don't think this is the correct toast that should be shown in a receiver device. And whatever be the status of the forms (already exists, new form downloaded, etc) it should be shown in the dialog box with each of the form transfer result, and this should happen in both sender and receiver device. Currently, in the sender's device, the results are printed on the activity itself.
I'm not sure the issue is with this PR or with the master as well. @huangyz0918 Can you take a look at it.

huangyz0918:
Yes @lakshyagupta21 , this issue I have noticed already, the result should shown in the result dialog rather than print in the activity, this is not about the branch, and happens in master as well. So I plan to remove all the toasts and build a result dialog for bluetooth just the same as hotspot. We can file a new issue for that.

huangyz0918:
Closes #293

lakshyagupta21:
> Closes #293

@huangyz0918 It would be easy to track if you mention it in the PR description, also please mark this PR as `WIP`. If you're fixing the other issue in the same PR.

huangyz0918:
This PR is ready for a review, @lakshyagupta21 , please give some suggestions that can help to improve it.

lakshyagupta21:
The above two are sender and receiver and the status present in dialog is not consistent, I mean they should be printed in same format. And there should be enough spacing between two items.
You can refer the Collect, how we print there.
For reference I've attached an image.

huangyz0918:
> The above two are sender and receiver and the status present in dialog is not consistent, I mean they should be printed in same format. And there should be enough spacing between two items.
You can refer the Collect, how we print there.

@lakshyagupta21 Fixed and improved.

lakshyagupta21:
@huangyz0918 Can you add Version and Id in `()` and add form status i.e., `Blank form received` and other statuses in `[]`

huangyz0918:
> Can you add Version and Id in () and add form status i.e., Blank form received and other statuses in []

Ok, let me think which form is better. @lakshyagupta21 , will update soon.

lakshyagupta21:
There is one more issue I faced but I'm not sure is it reproducable everytime or not
Whenever I clean install the app and I go to receiver activity of bluetooth it asks me for location permission and I allow it but after scanning is completed it asked me to enable it from settings when I go there and do nothing and again come back everything works fine. I think there is some issue  with dialog show and dismiss.

huangyz0918:
All right, I'll take more tests today, is this issue appears in `master` ? @lakshyagupta21 I think it may related to PR #286 .

lakshyagupta21:
I faced this issue in this branch, not sure about master we've to check in master.

huangyz0918:
Cool, I'll do some tests, and send an issue if needed.

huangyz0918:
How about this form:

```
Form Name: XXXX
Version: XXXXX
ID: XXXX
Result: XXXX
----------------------
Form Name: XXXX
Version: XXXXX
ID: XXXX
Result: XXXX
----------------------
```

@lakshyagupta21 ?

lakshyagupta21:
> How about this form:
> 
> ```
> Form Name: XXXX
> Version: XXXXX
> ID: XXXX
> Result: XXXX
> ----------------------
> Form Name: XXXX
> Version: XXXXX
> ID: XXXX
> Result: XXXX
> ----------------------
> ```

This will be much more descriptive but yeah this also looks fine to me.

huangyz0918:
Okey, updated, you can have a quick look at this screenshot.

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtReceiverActivity.java
```diff
@@ -210,9 +238,6 @@ private Disposable addDownloadEventSubscription() {
                 .observeOn(schedulerProvider.androidThread())
                 .subscribe(downloadEvent -> {
                     switch (downloadEvent.getStatus()) {
-                        case QUEUED:
```
lakshyagupta21: Why this is removed?

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtReceiverActivity.java
```diff
@@ -222,14 +247,16 @@ private Disposable addDownloadEventSubscription() {
...
                         case FINISHED:
                             progressDialog.dismiss();
-                            Toast.makeText(this, getString(R.string.tv_form_send_success), Toast.LENGTH_SHORT).show();
+                            String result = downloadEvent.getResult();
+                            resultDialog.setMessage(result);
+                            resultDialog.show();
                             break;
                         case ERROR:
-                            progressDialog.dismiss();
+                            dismissAllDialogs();
```
lakshyagupta21: Ideally whichever dialog box is shown for showing the progress of form transfer, that should be dismissed here. Adding check for all the dialogs don't think its a good practice. Any reason for doing this way?

skunkworks_crow/src/main/java/org/odk/share/tasks/UploadJob.java
```diff
@@ -384,14 +392,14 @@ private void sendBlankForm(Cursor cursor) {
...
-                if (formVersion != null) {
-                    sbResult.append(getContext().getString(R.string.version, formVersion));
-                }
-                sbResult.append(getContext().getString(R.string.id, formId) + " " +
-                        getContext().getString(R.string.success, getContext().getString(R.string.blank_form_count,
-                                getContext().getString(R.string.sent))));
+                sbResult.append(getContext().getString(R.string.form_transfer_result,
+                        getContext().getString(R.string.success, ", " +
+                                getContext().getString(R.string.blank_form_count,
+                                        getContext().getString(R.string.received)))));
```
lakshyagupta21: Its a `UploadJob` so `R.string.sent` should be there instead of `R.string.received`

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtReceiverActivity.java
huangyz0918: I think the download job was queued is no need to let our user know, all they need to know is is transferring started and finished, what do you think? @lakshyagupta21

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtReceiverActivity.java
huangyz0918: I think that will  be much easier? 😢  All right, I'll fix this soon

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtReceiverActivity.java
lakshyagupta21: Few events are added for the understanding of developer, that's the only reason of adding it here, you can remove the toast message instead adding some log over there would be great

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtReceiverActivity.java
lakshyagupta21: I'm not saying that we can't do this, but if we go this way then we may end up in some situation where two dialogs are shown at a single point of time.
That's why handle the show and dismiss of dialogs where it is expected to be done rather than closing all at once.

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtReceiverActivity.java
huangyz0918: Great, will replace that using a timber log.

skunkworks_crow/src/main/java/org/odk/share/tasks/UploadJob.java
huangyz0918: Oops, thanks for reminding, will fix soon.

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtReceiverActivity.java
huangyz0918: I see, will improve.

huangyz0918: @lakshyagupta21 I keep this `dismissAll()` for `onStop()` method, and for `ERROR` case and `CANCEL` case, all we need to handle (dismiss) is progress dialog. Correct me if I'm wrong.

skunkworks_crow/src/main/java/org/odk/share/tasks/UploadJob.java
```diff
@@ -513,6 +521,12 @@ private void sendInstances(List<String> instanceIds, int progress, int total) {
         }
     }
 
+    private void setupResultTest(String formName, String status) {
+        sbResult.append(getContext().getString(R.string.form_name, formName) + "\n");
+        sbResult.append(getContext().getString(R.string.form_transfer_result, ", " + status));
```
lakshyagupta21: I think `,` should not be there as it will result into something like this
`Result: ,Success Sent for review`

huangyz0918: For string format, this case won't appears. `,` is always behind `Success`

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtReceiverActivity.java
lakshyagupta21: `onStop `will be called when the app goes in the background right? So what if the user goes in background and comes in the foreground, this will dismiss all the dialog box no matter what operation is happening right? That's the reason we need to handle it carefully, otherwise unexpected situations can arise.

skunkworks_crow/src/main/java/org/odk/share/tasks/UploadJob.java
huangyz0918: `R.string.form_transfer_result` is something like ` Success %s`.

lakshyagupta21: Nope, it's not the case can you check this once by transferring filled form.

lakshyagupta21: `R.string.form_transfer_result` is like `Result: %s` and `%s` is ` "," + status`

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtReceiverActivity.java
huangyz0918: When I write this code, I referred to `HpReceiverActivity.java`, that includes something like

```java
    @Override
    protected void onStop() {
        super.onStop();

        if (alertDialog != null && alertDialog.isShowing()) {
            alertDialog.dismiss();
        }
    }
```

I think we can just remove the `onStop()`, I'm wondering why we need to handle dialogs in `OnStop()`? @lakshyagupta21 right?

skunkworks_crow/src/main/java/org/odk/share/tasks/UploadJob.java
huangyz0918: Ohh, you are right, let me fix that.

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtReceiverActivity.java
lakshyagupta21: I don't remember completely why was it added there, thanks for pointing it out we've to do some tests for activity transitions for hotspot as well.

skunkworks_crow/src/main/java/org/odk/share/tasks/UploadJob.java
huangyz0918: updated, please review again

lakshyagupta21:
Everything looks good to me.
Can you take a look at the dialog issue?

huangyz0918:
> Can you take a look at the dialog issue?
Yeah, thanks for such detailed review, I'm working on the dialog issue now.

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtReceiverActivity.java
huangyz0918: I will remove the `onStop()` and the `dismissAll` method here since I think we have already handle dialogs in `addDownloadEventSubscription()`, and we can check if it is also necessary in hotspot receiver and improve that.

lakshyagupta21: +1

huangyz0918: improved.

lakshyagupta21:
There is an issue that I'm facing in this PR
Problem description:
When you tap on `Receive` and select Bluetooth, it opens Bluetooth Receiver activity and asks the user for location permission as well as show the dialog behind the `System dialog asking for permission` and when you allow the location permission the dialog is shown asking for location permission again. Ideally, the dialog should not be shown until the permission is denied.

lakshyagupta21:
Is this issue resolve in #270 ?

huangyz0918:
> When you tap on Receive and select Bluetooth, it opens Bluetooth Receiver activity and asks the user for location permission as well as show the dialog behind the System dialog asking for permission and when you allow the location permission the dialog is shown asking for location permission again. Ideally, the dialog should not be shown until the permission is denied.

I cannot reproduce this issue in both master and this branch.

> Is this issue resolve in #270 ?

No

PR #286, opened 2019-07-28
[MRG] Add the location permission for android 9.

### LINKED ISSUE #284: Location permission required in android 9
I tested the bluetooth feature in PR #272 .

#### Problem description
In the android 9 devices, we need to grant the location for the bluetooth scanning. Only the location permission granted, the bluetooth can discover other devices. In other APIs, this issue won't appear. 

- In android 9 device
- Start the receiver and scanning devices using bluetooth.

#### Expected behavior
We should request for the location permission if our users are trying to start the bluetooth. 

#### Other information 
@opendatakit-bot claim

huangyz0918:
Closes #284 

Tested in my Nexus 6P (android 9.0) with success.

This fix the android 9 permission issue with bluetooth, and after this issue was resolved, the bluetooth feature can work well in android 9 devices I think. 

N/A

### DIFF
--- a/skunkworks_crow/build.gradle
+++ b/skunkworks_crow/build.gradle
@@ -70,4 +70,8 @@ dependencies {
 
     //MPAndroidChart
     implementation 'com.github.PhilJay:MPAndroidChart:v3.0.3'
+
+    // Permissions Dispatcher
+    implementation "org.permissionsdispatcher:permissionsdispatcher:4.5.0"
+    annotationProcessor "org.permissionsdispatcher:permissionsdispatcher-processor:4.5.0"
 }
--- a/skunkworks_crow/src/main/assets/open_source_licenses.html
+++ b/skunkworks_crow/src/main/assets/open_source_licenses.html
@@ -29,6 +29,15 @@ limitations under the License.
 </h3>
 <ul>
     <li>
+        <b>Permissions Dispatcher</b>
+    </li>
+    <li>
+        <b>Mockito</b>
+    </li>
+    <li>
+        <b>Robolectric</b>
+    </li>
+    <li>
         <b>Dagger 2</b>
     </li>
     <li>
--- a/skunkworks_crow/src/main/assets/open_source_licenses.html
+++ b/skunkworks_crow/src/main/assets/open_source_licenses.html
@@ -29,7 +29,7 @@ limitations under the License.
 </h3>
 <ul>
     <li>
-        <b>Permissions Dispatcher</b>
+        <b>PermissionsDispatcher</b>
     </li>
     <li>
         <b>Dagger 2</b>
--- a/skunkworks_crow/src/main/assets/open_source_licenses.html
+++ b/skunkworks_crow/src/main/assets/open_source_licenses.html
@@ -88,5 +88,24 @@ http://www.apache.org/licenses/LICENSE-2.0
 Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
 </pre>
 
+<h3>
+    The MIT License (MIT)
+</h3>
+<ul>
+    <li>
+        <b>Mockito</b>
+    </li>
+    <li>
+        <b>Robolectric</b>
+    </li>
+</ul>
+<pre>
+Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+</pre>
+
 </body>
 </html>
--- a/skunkworks_crow/src/main/assets/open_source_licenses.html
+++ b/skunkworks_crow/src/main/assets/open_source_licenses.html
@@ -32,12 +32,6 @@ limitations under the License.
         <b>Permissions Dispatcher</b>
     </li>
     <li>
-        <b>Mockito</b>
-    </li>
-    <li>
-        <b>Robolectric</b>
-    </li>
-    <li>
         <b>Dagger 2</b>
     </li>
     <li>
--- a/skunkworks_crow/src/main/java/org/odk/share/utilities/PermissionUtils.java
+++ b/skunkworks_crow/src/main/java/org/odk/share/utilities/PermissionUtils.java
@@ -1,14 +1,18 @@
 package org.odk.share.utilities;
 
 
+import android.Manifest;
 import android.app.Activity;
 import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
+import android.content.pm.PackageManager;
 import android.location.LocationManager;
 import android.provider.Settings;
 
 import androidx.appcompat.app.AlertDialog;
+import androidx.core.app.ActivityCompat;
+import androidx.core.content.ContextCompat;
 
 import org.odk.share.R;
 
@@ -19,6 +23,9 @@ import static android.content.Context.LOCATION_SERVICE;
  * For putting utils for permission checks.
  */
 public class PermissionUtils {
+
+    private static final int PERMISSIONS_REQUEST_FINE_LOCATION = 0x110;
+    private static final int PERMISSIONS_REQUEST_COARSE_LOCATION = 0x111;
 
     private PermissionUtils() {
     }
@@ -55,6 +62,31 @@ public class PermissionUtils {
 
         builder.setCancelable(false);
         builder.show();
+    }
+
+    /**
+     * Request for the location permission.
+     */
+    public static void requestLocationPermission(Activity activity) {
+        // ACCESS_FINE_LOCATION
+        if (ContextCompat.checkSelfPermission(activity.getApplicationContext(), Manifest.permission.ACCESS_FINE_LOCATION)
+                != PackageManager.PERMISSION_GRANTED) {
+            ActivityCompat.requestPermissions(activity,
+                    new String[]{Manifest.permission.ACCESS_FINE_LOCATION},
+                    PERMISSIONS_REQUEST_FINE_LOCATION);
+            ActivityCompat.shouldShowRequestPermissionRationale(activity,
+                    Manifest.permission.ACCESS_FINE_LOCATION);
+        }
+
+        // ACCESS_COARSE_LOCATION
+        if (ContextCompat.checkSelfPermission(activity.getApplicationContext(), Manifest.permission.ACCESS_COARSE_LOCATION)
+                != PackageManager.PERMISSION_GRANTED) {
+            ActivityCompat.requestPermissions(activity,
+                    new String[]{Manifest.permission.ACCESS_COARSE_LOCATION},
+                    PERMISSIONS_REQUEST_COARSE_LOCATION);
+            ActivityCompat.shouldShowRequestPermissionRationale(activity,
+                    Manifest.permission.ACCESS_COARSE_LOCATION);
+        }
     }
 }
 
--- a/skunkworks_crow/src/main/java/org/odk/share/utilities/PermissionUtils.java
+++ b/skunkworks_crow/src/main/java/org/odk/share/utilities/PermissionUtils.java
@@ -6,6 +6,7 @@ import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
 import android.location.LocationManager;
+import android.net.Uri;
 import android.provider.Settings;
 
 import androidx.appcompat.app.AlertDialog;
@@ -19,6 +20,8 @@ import static android.content.Context.LOCATION_SERVICE;
  * For putting utils for permission checks.
  */
 public class PermissionUtils {
+
+    private static final String SCHEME = "package";
 
     private PermissionUtils() {
     }
@@ -45,6 +48,30 @@ public class PermissionUtils {
 
         builder.setPositiveButton(targetActivity.getString(R.string.settings), (DialogInterface dialog, int which) -> {
             Intent intent = new Intent(Settings.ACTION_LOCATION_SOURCE_SETTINGS);
+            targetActivity.startActivity(intent);
+        });
+
+        builder.setNegativeButton(targetActivity.getString(R.string.cancel), (DialogInterface dialog, int which) -> {
+            dialog.dismiss();
+            targetActivity.finish();
+        });
+
+        builder.setCancelable(false);
+        builder.show();
+    }
+
+    /**
+     * Showing an alert dialog and send users to app info page.
+     */
+    public static void showAppInfo(Activity targetActivity, String packageName) {
+        AlertDialog.Builder builder = new AlertDialog.Builder(targetActivity);
+        builder.setMessage(R.string.permission_open_info);
+
+        builder.setPositiveButton(targetActivity.getString(R.string.permission_open_info_button), (DialogInterface dialog, int which) -> {
+            Intent intent = new Intent();
+            intent.setAction(Settings.ACTION_APPLICATION_DETAILS_SETTINGS);
+            Uri uri = Uri.fromParts(SCHEME, packageName, null);
+            intent.setData(uri);
             targetActivity.startActivity(intent);
         });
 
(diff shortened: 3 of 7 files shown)


skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtReceiverActivity.java
```diff
@@ -78,6 +79,9 @@ protected void onCreate(Bundle savedInstanceState) {
         setTitle(getString(R.string.connect_bluetooth_title));
         setSupportActionBar(toolbar);
 
+        //checking the location permission for bluetooth for android 9.
+        PermissionUtils.requestLocationPermission(this);
```
lakshyagupta21: If it's required for only Android 9 devices, then I think we should not ask for this permission unnecessarily for other devices in which it is not required.

huangyz0918: Other devices don't need to ask for permission because it was granted by default but we don't know if there are something unexpected. The code has `if-else` to check so I think this more insurance, what do you think? @lakshyagupta21

huangyz0918: I can remove that comments.

lakshyagupta21: How the permissions are granted by default, I'm asking this because `Location` is considered as [dangerous permission](https://developer.android.com/guide/topics/permissions/overview#permission-groups) and for devices greater than Android 5 apps should ask this at runtime to user, otherwise app doesn't have this permission.


> Other devices don't need to ask for permission

Are you talking about devices with versions lower than 5?

huangyz0918: Yes, but as for the devices in android 5-8, the application will give that permission without asking, this is why sometimes we don't need to ask for that. Anyway, in this change, the location permission is required for bluetooth, if the location didn't turned on, we can discover nothing. I added the `if` check to see if our application granted this permission, if not we will popup the dialog to request, so I think we don't have to set the specific API code (like target API is android 9), just call this `PermissionUtils.requestLocationPermission(this)`.

lakshyagupta21: I think you noticed that `Location` permission is required in Android 9 so that it can connect to bluetooth or to discover the bluetooth devices. Theoretically, app should not be granted `Dangerous` permission automatically if the device used is on version >= 5,0 and if the compileSDKVersion is also greater than or equal to 23. I just checked in Android 7.1 that it doesn't need any location permission to play with `Bluetooth`. And Bluetooth is also enabled even the location permission is not given to the app. 
Correct me if I'm wrong at any point.

huangyz0918: You are right, I'll update the code. Thanks for your explanation.

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtSenderActivity.java
```diff
@@ -73,6 +75,11 @@ protected void onCreate(Bundle savedInstanceState) {
         setTitle(getString(R.string.send_instance_title));
         setSupportActionBar(toolbar);
 
+        //checking the location permission for bluetooth for android 9.
```
lakshyagupta21: I think it would be better to ask for this permission before opening the activity(When bluetooth option is choosen). Because both the permission asking dialog and scanning dialog appears at the same time .

huangyz0918: Can we grant this permission when we are entering the main page? If we choose bluetooth then popup the permission dialog, that page still has the method choosing dialog appears at the same time.

lakshyagupta21: If we ask for location permission on start, then what if the user never chooses `Bluetooth` as the transfer mode then it would not be of no use. 
I'm suggesting to open a dialog asking for user permission when user taps on `Bluetooth` as the transfer mode(Close the dialog asking for the mode of transfer then only show the dialog for the permission) and if user allows then send to BluetoothReceiver/Sender Screen otherwise again show the dialog saying that `Location permission is needed for Bluetooth transfer` with two button options `Cancel`(negative) and `Access permission`(positive). `Access permission` will ask user again for the permission and if he has selected the checkbox `Never ask again` when permission was asked then you won't be able to request for permission again, in that case, you have to navigate user to our apps' settings so that he can manually go and give permission to us.
I know its pretty long process and but I think this is the way to tackle this issue.

lakshyagupta21: This is how generally other apps tackle this, in case you've some other option in mind please let me know.

huangyz0918: Cool, I'll follow your suggestion. But can I use some run-time android permission library? So that I can remove the `PermissionUtil.java` and make the code more flexible, something like https://github.com/Karumi/Dexter, it can also help me to build a dialog.

lakshyagupta21: Sure, It totally depends on you what you want to use, and but don't forget to add it in our `Open Source Licenses` section, its always to good to give the **credits**. And yes if you're doing it for Location permission then do it for every dangerous permission like `read and write storage`. Or you can open a different PR for that.

huangyz0918:
@lakshyagupta21 I got a better idea, you can have a review of the code now.

lakshyagupta21:
This also looks good to me.

skunkworks_crow/src/main/res/values/strings.xml
```diff
@@ -157,4 +157,8 @@
     <string name="dev_already_connected">already connected to the device, start transferring now&#8230;</string>
     <string name="tv_form_send_success">form sent with success! \n</string>
     <string name="bluetooth_no_device_scanned">No devices are available. \n Tap refresh to search again.</string>
+    <string name="button_allow">Allow</string>
+    <string name="button_deny">Deny</string>
+    <string name="permission_location_rationale">In Android 9, we need the location permission to make the bluetooth work, please give us the location permission and move forward.</string>
```
lakshyagupta21: I don't think that its a good idea to mention any device issues to the user. Lets just ask him for the permission.

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtSenderActivity.java
```diff
@@ -148,18 +158,81 @@ private Disposable addUploadEventSubscription() {
...
+        Toast.makeText(this, R.string.permission_location_never_ask, Toast.LENGTH_SHORT).show();
+        finish();
+    }
+
+    /**
+     * If clicked the "never ask", we should show a toast to guide user.
+     */
+    @OnNeverAskAgain({Manifest.permission.ACCESS_FINE_LOCATION,
+            Manifest.permission.ACCESS_COARSE_LOCATION})
+    void showNeverAskForCamera() {
```
lakshyagupta21: `showNeverAskForCamera` method name doesn't match with the permission that we're requesting for.

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtSenderActivity.java
```diff
@@ -148,18 +158,81 @@ private Disposable addUploadEventSubscription() {
...
+        Toast.makeText(this, R.string.permission_location_never_ask, Toast.LENGTH_SHORT).show();
```
lakshyagupta21: Rather than printing toast, it would be helpful for a user if we show some dialog where he can choose to navigate to settings directly. Otherwise, he may have to do a lot of efforts in searching the app in settings and then enable the location.

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtSenderActivity.java
```diff
@@ -148,18 +158,81 @@ private Disposable addUploadEventSubscription() {
...
+                .show();
+    }
+
+    /**
+     * If the permission was denied, finishing this activity.
+     */
+    @OnPermissionDenied({Manifest.permission.ACCESS_FINE_LOCATION,
+            Manifest.permission.ACCESS_COARSE_LOCATION})
+    void showDeniedForLocation() {
+        Toast.makeText(this, R.string.permission_location_never_ask, Toast.LENGTH_SHORT).show();
```
lakshyagupta21: +1

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtReceiverActivity.java
```diff
@@ -106,33 +135,75 @@ private void setupScanningDialog() {
...
+        Toast.makeText(this, R.string.permission_location_never_ask, Toast.LENGTH_SHORT).show();
+        finish();
+    }
+
+    /**
+     * If clicked the "never ask", we should show a toast to guide user.
+     */
+    @OnNeverAskAgain({Manifest.permission.ACCESS_FINE_LOCATION,
+            Manifest.permission.ACCESS_COARSE_LOCATION})
+    void showNeverAskForCamera() {
```
lakshyagupta21: Method name doesn't match with the permissions requested.

huangyz0918:
@lakshyagupta21 thanks for your detailed review! I think the code is ready now.

skunkworks_crow/build.gradle
```diff
@@ -70,4 +70,8 @@ dependencies {
 
     //MPAndroidChart
     implementation 'com.github.PhilJay:MPAndroidChart:v3.0.3'
+
+    // Permissions Dispatcher
+    implementation "org.permissionsdispatcher:permissionsdispatcher:4.5.0"
```
lakshyagupta21: Also add about this library in our `Open source license` section.

huangyz0918: Oooh yes, thanks for reminding, wait a minute.

huangyz0918: Updated along with `Robolectric` and `Mockito`.

skunkworks_crow/src/main/assets/open_source_licenses.html
```diff
@@ -28,6 +28,15 @@ <h3>
     Apache License Version 2.0, January 2004
 </h3>
 <ul>
+    <li>
+        <b>Permissions Dispatcher</b>
+    </li>
+    <li>
+        <b>Mockito</b>
```
lakshyagupta21: Mockito and Roboelectric are MIT licensed they should not come under Apache section.
You can refer https://github.com/opendatakit/collect/blob/master/collect_app/src/main/assets/open_source_licenses.html for adding the MIT License section

huangyz0918:
I think it's ready for a review @lakshyagupta21 .

Chromicle:
@huangyz0918 I just checked in android 6(Tablet) I am unable to send the forms it is coming back 
in logs, I can't see any crash message or error message I don't understood why it happens

huangyz0918:
@Chromicle This PR is not related to android 6, can you work normally in master?

skunkworks_crow/src/main/res/values/strings.xml
```diff
@@ -157,4 +157,7 @@
     <string name="dev_already_connected">already connected to the device, start transferring now&#8230;</string>
     <string name="tv_form_send_success">form sent with success! \n</string>
     <string name="bluetooth_no_device_scanned">No devices are available. \n Tap refresh to search again.</string>
+    <string name="permission_location_denied">You need to get location permission to enable bluetooth feature!</string>
+    <string name="permission_open_info">Please get location permissions on the APP details page.</string>
```
lakshyagupta21: Change text to `Skunkworks-crow app requires location permission to access your Bluetooth. Please allow location permission from your app settings.`

skunkworks_crow/src/main/res/values/strings.xml
```diff
@@ -157,4 +157,7 @@
...
+    <string name="permission_open_info_button">Details</string>
```
lakshyagupta21: Change it to `Open Settings`

skunkworks_crow/src/main/java/org/odk/share/utilities/PermissionUtils.java
```diff
@@ -56,6 +59,30 @@ public static void showLocationAlertDialog(Activity targetActivity) {
...
+    public static void showAppInfo(Activity targetActivity, String packageName) {
+        AlertDialog.Builder builder = new AlertDialog.Builder(targetActivity);
+        builder.setMessage(R.string.permission_open_info);
+
+        builder.setPositiveButton(targetActivity.getString(R.string.permission_open_info_button), (DialogInterface dialog, int which) -> {
+            Intent intent = new Intent();
+            intent.setAction(Settings.ACTION_APPLICATION_DETAILS_SETTINGS);
+            Uri uri = Uri.fromParts(SCHEME, packageName, null);
+            intent.setData(uri);
+            targetActivity.startActivity(intent);
```
lakshyagupta21: There is an issue over here when the user moves to app settings, and no matter what he does there and after doing all the activity on app settings when he presses back he goes back to the `Send data` screen with empty view.
You should check whenever an activity is resumed, whether user actually left application to actually play with the app settings, if the answer is yes then check if we've enough permission to do the transmission, if not then again display the same dialog.

skunkworks_crow/src/main/res/values/strings.xml
```diff
@@ -157,4 +157,7 @@
     <string name="dev_already_connected">already connected to the device, start transferring now&#8230;</string>
     <string name="tv_form_send_success">form sent with success! \n</string>
     <string name="bluetooth_no_device_scanned">No devices are available. \n Tap refresh to search again.</string>
+    <string name="permission_location_denied">You need to get location permission to enable bluetooth feature!</string>
```
lakshyagupta21: Change it to `Location permission is required to send the forms using Bluetooth. Please allow it in the app settings.`

skunkworks_crow/src/main/java/org/odk/share/utilities/PermissionUtils.java
huangyz0918: Fixed, I tested in android 9 and non-android 9 devices, no issues, but I think it still needs more tests.

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtReceiverActivity.java
```diff
@@ -106,33 +137,60 @@ private void setupScanningDialog() {
...
+        }
+    }
+
+    /**
+     * If the permission was denied, finishing this activity.
+     */
+    @OnPermissionDenied({Manifest.permission.ACCESS_FINE_LOCATION,
+            Manifest.permission.ACCESS_COARSE_LOCATION})
+    void showDeniedForLocation() {
+        Toast.makeText(this, R.string.permission_location_denied, Toast.LENGTH_SHORT).show();
```
lakshyagupta21: Change the toast length to `LONG`, everywhere

lakshyagupta21:
I think we've to enable location permission for all device because in Android 7.1 I've to go and enable the location permission from the settings then only it starts showing the bluetooth devices when I start scanning it.

huangyz0918:
> I think we've to enable location permission for all device because in Android 7.1 I've to go and enable the location permission from the settings then only it starts showing the bluetooth devices when I start scanning it.

So you mean I should remove the android 9 checks?

lakshyagupta21:
Can you check this once in any of your Android 6 and Android 7 device, don't forget to check for location permission (verify when enabled/disabled)

huangyz0918:
Ok, will do that and update this PR.

huangyz0918:
> Can you check this once in any of your Android 6 and Android 7 device, don't forget to check for location permission (verify when enabled/disabled)

You can checkout [this](https://stackoverflow.com/questions/38973118/which-permission-do-we-need-at-runtime-in-android-6-0), from Android 6, the `ACCESS_COARSE_LOCATION` and `ACCESS_FINE_LOCATION` need to be obtained from run time. So I think we should remove the API checks and apply that to all devices. @lakshyagupta21 I updated this PR and please review.

lakshyagupta21:
> `ACCESS_COARSE_LOCATION` and `ACCESS_FINE_LOCATION` need to be obtained from run time

I know that these permissions are required to be asked at runtime, but my doubt was why do we need to ask it I mean who requires this permission.
We found this problem for Android 9, then how was it working previously for other versions if location permission was not given to any of them.

huangyz0918:
> I know that these permissions are required to be asked at runtime, but my doubt was why do we need to ask it I mean who requires this permission.

I have done some tests on different devices using both real devices and virtual devices.

|Device Name|Android Version|Needs Location Permission for Bluetooth|
|:--:|:--:|:--:|
|Nexus 6P|9.0.0|Yes|
|Huawei P10|7.1.0|Yes|
|Nexus 5X|6.0.1|Yes|
|Samsung Galaxy Note3|5.0.0|No|

The conclusion is, in versions of Android prior to Marshmallow, the user could use Bluetooth without location services enabled, we need the location permission in all the devices with an API level > 6.0. @lakshyagupta21

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtSenderActivity.java
```diff
@@ -148,18 +159,68 @@ private Disposable addUploadEventSubscription() {
...
+        switch (requestCode) {
+            case DISCOVERABLE_CODE:
+                if (resultCode == SUCCESS_CODE) {
+                    startCheckingDiscoverableDuration();
+                } else {
+                    finish();
+                }
+                break;
+            case APP_SETTING_REQUEST_CODE:
+                //only android 9 would enter this case.
```
lakshyagupta21: @huangyz0918 Please remove this comment, now every device with API >= 21 will execute this block.

huangyz0918: Sure! do this right now.

PR #290, opened 2019-08-03
Default transferring mechanism and UI improvements

### LINKED ISSUE #266: [Feature] Build the default transferring mechanism
#### Problem description
In #260 , we added a new approach using bluetooth. 
Currently, we will ask users to choose a transfer method every time when they are trying to sen/receive forms. Should we ask them every time? Can we set a default method and auto switch to another method if the default failed? 

According to my proposal, we can set the default method in our system settings, so we need to refactor the method choosing dialog, and we can build a better UI/UX based on this refactor.

What do you think about that? @lakshyagupta21 @shobhitagarwal1612 @Chromicle 
#### Expected behavior

We shall have a default method, and if the default method has any issue or crashed while transferring data, we can auto switch to the other one (bluetooth/hotspot).

huangyz0918:
Closes #266 

I tested this PR in my Nexus 6P (android 9) and Samsung Galaxy Note 3 (android 5).

This PR is still work in progress, there are some tasks we need to do:
- [x] add a default option in preference page.
- [x] remove the dialog that asking for choosing methods.
- [x] add the notification to ask if user need to switch method.
- [x] improve the UI and make the bluetooth and the hotspot different.

Extra improvements:
- [x] add a common method for present the results in a dialog. (finish the bluetooth activity after transmission completed)
- [x] add some animation in the bluetooth sending and receiving pages, make a better UI.


skunkworks_crow/build.gradle
```diff
@@ -70,4 +70,7 @@ dependencies {
 
     //MPAndroidChart
     implementation 'com.github.PhilJay:MPAndroidChart:v3.0.3'
+
+    //Loading Animations
+    implementation 'com.github.ybq:Android-SpinKit:1.2.0'
```
lakshyagupta21: I would say not to use any animation library till we don't have a proper need or requirement for it. Even in ODK-Collect we don't have any animation library so let's make it consistent everywhere.

huangyz0918: Sure!

huangyz0918:
Note, I have done crazy tests in the switching method option, and commit [3d69968](https://github.com/opendatakit/skunkworks-crow/pull/290/commits/3d69968e08059711f517b6ecab62d9d4cf0e358f) still has some issues, the biggest issue that when I switched the method from bluetooth -> hotspot -> bluetooth -> hotspot -> bluetooth... and so on, there will be an exception like `socked closed`, I think this related to the socket handling, and should be review detailed here.

huangyz0918:
@lakshyagupta21 I changed the colors in `BtSenderActivity.java` and `BtReceiverActivity.java` to blue and leave other pages the same as the original, what do you think about these changes? 

I have three proposals
- A blue color for `BtSenderActivity.java` and `BtReceiverActivity.java`, green for others.
- A blue color for `BtSenderActivity.java` and `BtReceiverActivity.java`, green for `HpSenderActivity.java` and `HpReceiverActivity.java`, and another color for other pages (`MainActivity.java`, etc).
- Apply the different color to all pages according `default methods` in settings.

Which one is better? @lakshyagupta21 @shobhitagarwal1612 

Blue color looks like:

lakshyagupta21:
I don't think the theme will help that much, because only toolbar color looks changes, which doesn't catch the user's attention much.
To catch user's attention either the whole screen should be colored (which I'm not sure whether that will look good or not)  or adding some icon (bluetooth or hotspot) with different color.

huangyz0918:
> I don't think the theme will help that much, because only toolbar color looks changes, which doesn't catch the user's attention much.

I changed the `colorPrimary` and some other colors so the button color even the progress bar color will change. Yeah, if it's not enough, I will add the icons and let's discuss more about how to make that looks better.

lakshyagupta21:
> I changed the `colorPrimary` and some other colors so the button color even the progress bar color

A progress bar and buttons would be shown only when the transmission is started or when its scanning for other devices, right? So I think the first step is to identify which transfer mechanism is chosen by sender how will you identify at receiver that what mechanism sender has chosen?

huangyz0918:
> how will you identify at receiver that what mechanism sender has chosen?
Our receiver cannot know which method the sender selected. But we can build a discriminative UI that can tell users the methods are different. So what's your opinion about building that? I have three proposals [above](https://github.com/opendatakit/skunkworks-crow/pull/290#issuecomment-519758821) and what do you think about them? @lakshyagupta21

huangyz0918:
@opendatakit-bot label "needs review"

lakshyagupta21:
@huangyz0918 Transfer is not working when bluetooth option is selected. It keeps shows `Connecting` state.



> Our receiver cannot know which method the sender selected. But we can build a discriminative UI that can tell users the methods are different. So what's your opinion about building that? I have three proposals [above](https://github.com/opendatakit/skunkworks-crow/pull/290#issuecomment-519758821) and what do you think about them? @lakshyagupta21

We should not change any theme or activity color, that should be kept the same for all activity. 
As much I remember in our call last time we all agreed to add some icons on the screen with different color, For BlueTooth on both the screens (receiver's and sender's) blue color bluetooth icons should be there and some other different color(dark) for a hotspot. Or maybe we can do one thing change the color of the whole screen (to blue for bluetooth and some other dark color for hotspot) and add the white color UI elements.

huangyz0918:
> add some icons on the screen with different color

I see, so where should we place the icons? @lakshyagupta21 

In the tool bar? I'm afraid only icons in the tool bar is not so obvious, like these:

|||
|:--:|:--:|
|||

lakshyagupta21:
@huangyz0918 Dialog is shown at the second step, we're trying to solve problem when user is on the screen and doesn't know about the transferring medium chosen, and only changing the theme won't work.

Here are few mocks I've created to convey our idea in better way, so that you can understand the requirement better. 

Screen1
Screen2

Colors and text can be changed at a later stage.

Maybe @shobhitagarwal1612 @lognaturel @yanokwa Can give more suggestions to your ideas.

huangyz0918:
> Transfer is not working when bluetooth option is selected. It keeps shows Connecting state.

Can you offer some more information like how to reproduce and error logs?  @lakshyagupta21 
 Thanks! I'll try to fix issues as soon as possible.

huangyz0918:
> Here are few mocks I've created to convey our idea in better way, so that you can understand the requirement better.

So do we need to reactor the UI for hotspot pages? @lakshyagupta21 I think there are many components already in that pages (like there are already a QR code for hotspot sender, no much room for a big logo, and that will make bluetooth and hotspot sender inconsistent), it's hard for us to build a page for hotspot like you proposed. As for me, I think adding an icon in the title bar is enough for user to know which method he selected. What do you think? 

@shobhitagarwal1612 @lognaturel @yanokwa

lakshyagupta21:
> As for me, I think adding an icon in the title bar is enough for the user to know which method he selected.

I'm not sure adding an icon in the toolbar just to catch the user attention is a good android design practice, but let's see what others have to say about it. Maybe in future, we may come up with QR Code feature for Bluetooth as well so we've to consider that as well.

huangyz0918:
Cool, let see what others said @lakshyagupta21 .
For receivers, both bluetooth and hotspot have a recycler view for many items (devices list for bluetooth and wifi list for hotspot, it's also an issue is we want to add a big icon in the page), I don't know how to make the methods more obvious for our users, maybe change color for the whole page? I'm not sure that is a pretty look.

lakshyagupta21:
> For receivers, both bluetooth and hotspot have a recycler view for many items (devices list for bluetooth and wifi list for a hotspot, it's also an issue if we want to add a big icon in the page)

We won't be adding an icon to list page, that page is dedicated to select the device with whom you want to make the connection, the icons and colors will be changed of only sender's screen. There is nothing that the sender can do by knowing which medium receiver has chosen, this feature is for the receiver to know that which medium sender has chosen so that he can switch the transfer medium

skunkworks_crow/src/main/java/org/odk/share/views/ui/hotspot/HpReceiverActivity.java
```diff
@@ -109,15 +114,19 @@
     private WifiConnector wifiConnector;
     private WifiStateBroadcastReceiver wifiStateBroadcastReceiver;
     private boolean isConnected = false;
+    private HpReceiverActivity thisActivity = this;
```
lakshyagupta21: Same as mentioned above, don't create separate variables

skunkworks_crow/src/main/java/org/odk/share/views/ui/hotspot/HpSenderActivity.java
```diff
@@ -137,6 +147,35 @@ public void onBackPressed() {
...
+    @Override
+    public boolean onCreateOptionsMenu(Menu menu) {
+        getMenuInflater().inflate(R.menu.switch_method_menu, menu);
+        final MenuItem switchItem = menu.findItem(R.id.menu_switch);
+
+        switchItem.setOnMenuItemClickListener(new MenuItem.OnMenuItemClickListener() {
+            @Override
+            public boolean onMenuItemClick(MenuItem item) {
+                DialogUtils.createMethodSwitchDialog(thisActivity, (DialogInterface dialog, int which) -> {
+                    Intent intent = receivedIntent;
```
lakshyagupta21: Use `receivedIntent` directly.

skunkworks_crow/src/main/res/drawable/ic_security_black_24dp.xml
```diff
@@ -0,0 +1,9 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
lakshyagupta21: If you've used any tool to generate these icons then please do mention about those, its always good practice to mention someone else's work if we're building over that.

skunkworks_crow/src/main/java/org/odk/share/views/ui/settings/SettingsActivity.java
```diff
@@ -56,26 +62,37 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+        btSecureModePreference = (CheckBoxPreference) findPreference(PreferenceKeys.KEY_BLUETOOTH_SECURE_MODE);
         odkDestinationDirPreference = (EditTextPreference) findPreference(PreferenceKeys.KEY_ODK_DESTINATION_DIR);
 
         prefs = PreferenceManager.getDefaultSharedPreferences(getApplicationContext());
 
+        defaultMethodPreference.setSummary(prefs.getString(PreferenceKeys.KEY_DEFAULT_TRANSFER_METHOD,
+                getString(R.string.hotspot)));
         hotspotNamePreference.setSummary(prefs.getString(PreferenceKeys.KEY_HOTSPOT_NAME,
                 getString(R.string.default_hotspot_ssid)));
+        bluetoothNamePreference.setSummary(prefs.getString(PreferenceKeys.KEY_BLUETOOTH_NAME,
```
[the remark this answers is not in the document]
huangyz0918: Fixed

skunkworks_crow/src/main/java/org/odk/share/utilities/DialogUtils.java
```diff
@@ -21,48 +26,77 @@ private DialogUtils() {
...
+        AlertDialog alertDialog = new AlertDialog.Builder(targetActivity).create();
+        alertDialog.setCancelable(false);
+        alertDialog.setButton(DialogInterface.BUTTON_POSITIVE, targetActivity.getString(R.string.cancel),
+                (dialog, i) -> {
+                    dialog.dismiss();
+                });
+
+        alertDialog.setButton(DialogInterface.BUTTON_NEGATIVE, targetActivity.getString(R.string.switch_method), listener);
+
+        if (targetActivity instanceof BtReceiverActivity || targetActivity instanceof BtSenderActivity) {
```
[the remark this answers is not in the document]
huangyz0918: No, the behavior is correct, maybe I should change to some more easy to understand string names.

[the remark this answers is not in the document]
huangyz0918: I mean, for example I need to switch to bluetooth from hotspot, what icon should we put in the dialog when user clicked the switch button? @lakshyagupta21

skunkworks_crow/src/main/java/org/odk/share/views/ui/bluetooth/BtSenderActivity.java
```diff
@@ -97,20 +106,24 @@ protected void onCreate(Bundle savedInstanceState) {
...
 
         setupDialog();
+        receivedIntent = getIntent();
         registerReceiver(bluetoothReceiver, new IntentFilter(BluetoothAdapter.ACTION_STATE_CHANGED));
 
         formIds = getIntent().getLongArrayExtra(INSTANCE_IDS);
         mode = getIntent().getIntExtra(MODE, ASK_REVIEW_MODE);
         if (formIds == null) {
-            formIds = getIntent().getLongArrayExtra(FORM_IDS);
+            formIds = receivedIntent.getLongArrayExtra(FORM_IDS);
```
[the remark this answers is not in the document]
huangyz0918: Added

skunkworks_crow/src/main/res/drawable/ic_security_black_24dp.xml
huangyz0918: Sure! These icons are generated by Android Studio, so I think we don't need to do that. But I'll mention other icons get from internet.

skunkworks_crow/src/main/java/org/odk/share/utilities/DialogUtils.java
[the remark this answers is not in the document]
lakshyagupta21: Okay, now I understand it could be more easy to understand if you write some more comments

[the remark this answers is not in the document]
huangyz0918: Comment added.

lakshyagupta21:
`Enable Secure Mode` doesn't seems to be working I've enabled that in sender's device but not in receiver's device but still it asked me to pair using code.
And @huangyz0918 now the UI is looking pretty amazing, thanks for making it good.

skunkworks_crow/src/main/java/org/odk/share/utilities/DialogUtils.java
```diff
@@ -21,48 +27,80 @@ private DialogUtils() {
...
-                                intent.setClass(ActivityUtils.getActivity(context), BtSenderActivity.class);
-                                break;
-                            case 1:
-                                intent.setClass(ActivityUtils.getActivity(context), HpSenderActivity.class);
-                                break;
-                        }
-                        context.startActivity(intent);
-                    }
-                }).create();
+    public static void switchToDefaultSendingMethod(@NonNull Context context, Intent intent) {
```
lakshyagupta21: Add `@NonNull` to Intent as well, because for that also we don't have any null check in method then expect it to be nonnull from the caller of this method

huangyz0918:
> doesn't seems to be working I've enabled that in sender's device but not in receiver's device but still it asked me to pair using code.

Yeah, the `Enable Secure Mode` was controlled by receiver, change settings in sender cannot affect the behavior. @lakshyagupta21

lakshyagupta21:
> Yeah, the `Enable Secure Mode` was controlled by receiver, change settings in sender cannot affect the behavior. @lakshyagupta21

Can you briefly mention that in the settings?

huangyz0918:
> Can you briefly mention that in the settings?

Sure

skunkworks_crow/src/main/res/values/strings.xml
```diff
@@ -139,15 +144,14 @@
...
-    <string name="method_bluetooth">Bluetooth</string>
-    <string name="method_wifi_hotspot">Hotspot</string>
-    <string name="title_send_options">Send Options</string>
-    <string name="title_receive_options">Receive Options</string>
     <string name="connect_bluetooth_title">Device to Receive</string>
     <string name="connected_bluetooth_downloading">Connected, start downloading&#8230;</string>
     <string name="connecting_title">Connecting</string>
     <string name="scanning_title">Scanning</string>
+    <string name="bluetooth">bluetooth</string>
+    <string name="hotspot">@string/default_hotspot_ssid</string>
```
lakshyagupta21: Why created different string resource? You can use `@string/default_hotspot_ssid` directly, right?

huangyz0918: Yes, I mean this will make the string name more make sense, right?

skunkworks_crow/src/main/java/org/odk/share/utilities/DialogUtils.java
huangyz0918: Sure, will do now.

skunkworks_crow/src/main/res/values/strings.xml
huangyz0918: Fixed

PR #296, opened 2019-08-11
[MRG] Fixed the storage permission handling progress.

### LINKED ISSUE #295: App not starting if permission is denied

#### Problem description
App doesn't start if application doesn't have storage permission.

1. Reset the app if already installed.
2. Install the app.
3. Deny the storage permission.
4. Open app again.
5. Mark the checkbox selected.
6. Again deny the storage permission.
7. Try opening the app again, it won't start.
#### Expected behavior
App should ask user to enable the permissions(through dialog box) from the app settings.

huangyz0918:
Closes #295 

I didn't use the permission dispatcher lib, since we already has the complete checking code, so I tried to fix this bug without a refactor. But I have tested with the permission dispatcher library as well, maybe something goes wrong with the injector, so the `OnPermissionDenied` would be called every time before I do any operations with the permission request dialog. So I just fixed the issue in the base of current code and make tests in my Nexus 6P (Android 9), everything works well.

N/A

N/A

### DIFF
--- a/skunkworks_crow/src/main/java/org/odk/share/views/ui/main/MainActivity.java
+++ b/skunkworks_crow/src/main/java/org/odk/share/views/ui/main/MainActivity.java
@@ -257,7 +257,6 @@ public class MainActivity extends FormListActivity implements LoaderManager.Load
                     setUpLoader();
                 } else {
                     if ((Build.VERSION.SDK_INT >= Build.VERSION_CODES.M &&
-                            checkSelfPermission(Manifest.permission.WRITE_EXTERNAL_STORAGE) != PackageManager.PERMISSION_GRANTED &&
                             !shouldShowRequestPermissionRationale(Manifest.permission.WRITE_EXTERNAL_STORAGE))) {
                         PermissionUtils.showAppInfo(this, getPackageName(), getString(R.string.permission_open_storage_info), getString(R.string.permission_storage_denied));
                     } else {
--- a/skunkworks_crow/src/main/java/org/odk/share/views/ui/main/MainActivity.java
+++ b/skunkworks_crow/src/main/java/org/odk/share/views/ui/main/MainActivity.java
@@ -251,35 +251,31 @@ public class MainActivity extends FormListActivity implements LoaderManager.Load
     @Override
     public void onRequestPermissionsResult(int requestCode, @NonNull String[] permissions, @NonNull int[] grantResults) {
         super.onRequestPermissionsResult(requestCode, permissions, grantResults);
-        switch (requestCode) {
-            case STORAGE_PERMISSION_REQUEST_CODE:
-                if (grantResults[0] == PackageManager.PERMISSION_GRANTED) {
-                    setUpLoader();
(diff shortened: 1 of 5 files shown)


skunkworks_crow/src/main/java/org/odk/share/views/ui/main/MainActivity.java
```diff
@@ -245,20 +251,40 @@ private void addListItemDivider() {
...
         super.onRequestPermissionsResult(requestCode, permissions, grantResults);
-
         switch (requestCode) {
             case STORAGE_PERMISSION_REQUEST_CODE:
                 if (grantResults[0] == PackageManager.PERMISSION_GRANTED) {
                     setUpLoader();
                 } else {
-                    //close the app if the permission is denied
-                    finish();
+                    if ((Build.VERSION.SDK_INT >= Build.VERSION_CODES.M &&
```
lakshyagupta21: I think if this block is executed that means permission is already denied, why're we checking again for the permission?

huangyz0918: Fixed

huangyz0918:
This PR can be review first. @lakshyagupta21 Thanks.

skunkworks_crow/src/main/java/org/odk/share/views/ui/main/MainActivity.java
```diff
@@ -245,20 +251,39 @@ private void addListItemDivider() {
...
```
lakshyagupta21: Even this version check is redundant because every time the version is going to be greater than `M`, because you request only for the greater versions than `M`

huangyz0918: But method `shouldShowRequestPermissionRationale` has a compiler check for API > 23, if we remove the API check, the code cannot pass the compiler.

skunkworks_crow/src/main/java/org/odk/share/views/ui/main/MainActivity.java
```diff
@@ -245,20 +251,39 @@ private void addListItemDivider() {
     @Override
     public void onRequestPermissionsResult(int requestCode, @NonNull String[] permissions, @NonNull int[] grantResults) {
         super.onRequestPermissionsResult(requestCode, permissions, grantResults);
-
         switch (requestCode) {
```
lakshyagupta21: And this switch case can be replaced with `if`, because switch is used when 3 or more items are present in the checklist.

huangyz0918: Sure, will improve soon.

skunkworks_crow/src/main/java/org/odk/share/views/ui/main/MainActivity.java
lakshyagupta21: Oops, I forgot that ok for now lets keep it.

Issue #197, opened 2019-03-10
Unable to send forms on android 8.0.0


Chromicle:
## Software and hardware versions
Android 8, Honor 8x 

### Problem description
Unable to send blank/filled forms on Android 8.0.0. The screen gets stuck on Waiting for the hotspot to connect
i.e QR code is not appearing to scan if we want to connect through wifi entering the password it is giving toast that you should scan a QR code to connect this network 

### Steps to reproduce the problem
Just select any blank or filled form and try to send on an android 9 device

### Expected behavior
It should work as expected(we can send the forms)

sarrocks1:
@Chromicle  Can you please give a detailed description . I have successfully run on honor 7x with android v8.0 and I m not facing any errors. Screenshots would help a lot.
Thank You.

iamvasuh:
@Chromicle is this issue still coming on your device?

Chromicle:
Yeah Iam facing that but not my device(friend) and I checked on realme 2 also it is still coming

PR #312, opened 2019-09-24
fix: multiple dialogs for location access

### LINKED ISSUE #311: bug: multiple dialogs appearing for requesting location access
Android v9.x.x, Vivo Z1 Pro

#### Problem description
When sending forms, the dialog for turn on the location is opening multiple times.
odk

Go to send blank form tab. Select the forms to send and click on the send button. If the location is not turned on, multiple dialogs will appear.

#### Expected behavior
A single dialog should be shown only.

naman653:
Closes #311 

Tested this on my Vivo Z1 Pro device running on Android v9.x.x

NA

It ensures better stability of app as multiple dialogs were using a lot of android resources which was slowing down the whole android framework.

### DIFF
--- a/skunkworks_crow/src/main/java/org/odk/share/views/ui/hotspot/HpSenderActivity.java
+++ b/skunkworks_crow/src/main/java/org/odk/share/views/ui/hotspot/HpSenderActivity.java
@@ -23,6 +23,7 @@ import androidx.annotation.NonNull;
 import androidx.annotation.RequiresApi;
 import androidx.appcompat.app.AlertDialog;
 import androidx.appcompat.widget.Toolbar;
+import androidx.core.app.ActivityCompat;
 
 import org.odk.share.R;
 import org.odk.share.events.HotspotEvent;
@@ -447,11 +448,25 @@ public class HpSenderActivity extends InjectableActivity {
     private void checkLocationPermission() {
 
         if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M ||
-                checkSelfPermission(Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED) {
+                checkSelfPermission(Manifest.permission.ACCESS_COARSE_LOCATION) == PackageManager.PERMISSION_GRANTED) {
 
             toggleHotspot();
         } else {
-            requestPermissions(new String[]{Manifest.permission.ACCESS_COARSE_LOCATION}, LOCATION_PERMISSION_REQUEST_CODE);
+            if (ActivityCompat.shouldShowRequestPermissionRationale(this,
+                    Manifest.permission.ACCESS_COARSE_LOCATION)) {
+
+                new AlertDialog.Builder(this)
+                        .setMessage(R.string.location_permission_needed)
+                        .setPositiveButton(R.string.ok, (dialogInterface, i) -> ActivityCompat.requestPermissions(this,
+                                new String[]{Manifest.permission.ACCESS_COARSE_LOCATION},
+                                LOCATION_PERMISSION_REQUEST_CODE))
+                        .create()
+                        .show();
+            } else {
+                ActivityCompat.requestPermissions(this,
+                        new String[]{Manifest.permission.ACCESS_COARSE_LOCATION},
+                        LOCATION_PERMISSION_REQUEST_CODE);
+            }
         }
     }
 


naman653:
@lakshyagupta21 Please review this.

skunkworks_crow/src/main/java/org/odk/share/views/ui/hotspot/HpSenderActivity.java
```diff
@@ -448,10 +449,26 @@ private void checkLocationPermission() {
 
         if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M ||
                 checkSelfPermission(Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED) {
-
             toggleHotspot();
         } else {
-            requestPermissions(new String[]{Manifest.permission.ACCESS_COARSE_LOCATION}, LOCATION_PERMISSION_REQUEST_CODE);
+            if (ActivityCompat.shouldShowRequestPermissionRationale(this,
+                    Manifest.permission.ACCESS_FINE_LOCATION)) {
```
huangyz0918: Where is the original `ACCESS_COARSE_LOCATION`?

naman653:
Thanks, @huangyz0918 for pointing out the issues. Please review now.

huangyz0918:
Looks good to me, let's ask @lakshyagupta21 for his opinion.

skunkworks_crow/src/main/java/org/odk/share/views/ui/hotspot/HpSenderActivity.java
```diff
@@ -447,11 +448,25 @@ private void createAlertDialog(String title, String message) {
     private void checkLocationPermission() {
 
         if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M ||
-                checkSelfPermission(Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED) {
+                checkSelfPermission(Manifest.permission.ACCESS_COARSE_LOCATION) == PackageManager.PERMISSION_GRANTED) {
```
lakshyagupta21: Why changing the fine location to coarse location, and have you tested transferring forms ater this change?

naman653: I understand what you are saying, but why we are using ACCESS_FINE_LOCATION in condition check but using ACCESS_COARSE_LOCATION while requesting permission.

lakshyagupta21: We're using Course Location at both the places, can you point me to the actual line where you have seen this issue?

naman653: I am saying why are we using FINE_LOCATION in the review you mentioned above, but COARSE_LOCATION at other places.

lakshyagupta21: There could be a miss while reviewing the code, as much as I remember when you ask for location permission it shows only one dialog and that allows both the permissions but if in our case if we need only one of the permission to execute the method or the piece of code that I think we should stick with the same permission not both of them. Thanks @naman653  for poinitng this out. It would be helpful if you can open a issue regarding this.

naman653: Happy to do that. But for which permission we should go for?

lakshyagupta21: That needs to be checked, if we're using both the permissions in our code then it's in our best interest to check for both of them while asking to the user.

naman653: Sorry for late replying, as I was busy with my mid terms. I think we should go for Coarse Location as it was being used before, and was working just fine before.

naman653: Also, tell me what should be done with this PR.

naman653: @lakshyagupta21 Please review

naman653:
@lakshyagupta21 Can you please review this or suggest someone who can. Seems like the community is not that active. It's been many days since I opened the PR. I want to work on more issues as there is so much that can be done on this application.

lakshyagupta21:
Thanks @naman653 for raising the PR,  this project is at its initial step we're still making it mature enough so that the features of this app doesn't break in field easily. So getting attention of lot of contributors will also take time,  currently me and @shobhitagarwal1612 are the ones who are involved in the PR reviews so I think he might be busy sonewhere in his work otherwise he would've reviewed and I'm travelling this week so I wasn't able to review. And btw no one is stopping you to work on other issues and PRs,  feel free to work on other issues,  we'll take a look at it soon.

lakshyagupta21:
@naman653 Sorry for late reply.
Code wise things looks good to me, I just wanted to check with you have you tried sending a form using hotspot method from one device to another? Its not working for me I'm trying to send from Android 7.1 to Android 9, it could be some other issue but just wanted to check if it's working for you or not.

naman653:
Yes, I was also unable to send forms. But I think it's another issue, as I was unable to send forms prior to the changes. There are older issues which points to the same problem #191, #197.

lakshyagupta21:
@naman653 , Thanks for taking a look into the issue and sorry for keeping this PR in review state for so long.
Merging this PR as the form issue is already there on the master, and I think we need to fix this asap otherwise it would create some problem while testing other issuees and PRs.

PR #314, opened 2019-10-07
fix: stop bluetooth after sharing forms according to initial state

### LINKED ISSUE #306: Stop turning off bluetooth after sharing forms when initial state is on
Android Samsung J8

#### Problem description
currently, after sending forms Bluetooth is turning off but when user is using Bluetooth before sharing forms but after sharing forms it is turning off so before turning off Bluetooth check initial state and if it is on leave that if it is off make it off

Share any form and check initial and after state

#### Expected behavior
if Bluetooth is on leave that if it is off make it off after sharing forms

naman653:
Closes #306 

Tested on my Vivo Z1 Pro running on Android 9.x.x

NA

Now if the Bluetooth is enabled before sharing forms, it will not be disabled after the process is completed.


naman653:
@lakshyagupta21 Please review

PR #338, opened 2020-01-11
fix: using app without odk collect installed

### LINKED ISSUE #337: bug: using app without installing ODK collect app
Android v9.x.x, Vivo Z1 Pro

#### Problem description
The current scenario of the app doesn't allow the user to use the app without ODK collect app. But one can access the app without installing the ODK collect app.
1. Open the app in absence of ODK collect app.
2. Click the install button.
3. Navigate back to the app without installing the ODK collect app.
4. Now the app is accessible without ODK collect installed.

#### Expected behavior
The app should not run until ODK collect is not installed.

naman653:
Closes #337 

Verified by testing on Vivo Z1 Pro running on Android v9.x.x
NA
It restricts the user from using the app without installing ODK collect app.

### DIFF
--- a/skunkworks_crow/src/main/java/org/odk/share/views/ui/main/MainActivity.java
+++ b/skunkworks_crow/src/main/java/org/odk/share/views/ui/main/MainActivity.java
@@ -60,6 +60,7 @@ public class MainActivity extends FormListActivity implements LoaderManager.Load
     private static final String FORM_CHOOSER_LIST_SORTING_ORDER = "formChooserListSortingOrder";
     private static final String COLLECT_PACKAGE = "org.odk.collect.android";
     private static final int STORAGE_PERMISSION_REQUEST_CODE = 101;
+    private static final int COLLECT_INSTALL_REQUEST_CODE = 102;
     private static final int FORM_LOADER = 2;
 
     @BindView(R.id.toolbar)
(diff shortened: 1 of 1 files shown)


huangyz0918:
Looks good to me, thanks for your contribution @naman653 , let @lakshyagupta21 to see if the code is ok to merge.

skunkworks_crow/src/main/java/org/odk/share/views/ui/main/MainActivity.java
```diff
@@ -213,9 +214,12 @@ private void showAlertDialog() {
 
         builder.setPositiveButton(getString(R.string.install), (DialogInterface dialog, int which) -> {
             try {
-                startActivity(new Intent(Intent.ACTION_VIEW, Uri.parse("market://details?id=" + COLLECT_PACKAGE)));
+                startActivityForResult(new Intent(Intent.ACTION_VIEW, Uri.parse("market://details?id=" + COLLECT_PACKAGE)),
+                        COLLECT_INSTALL_REQUEST_CODE);
             } catch (ActivityNotFoundException e) {
-                startActivity(new Intent(Intent.ACTION_VIEW, Uri.parse("https://play.google.com/store/apps/details?id=" + COLLECT_PACKAGE)));
+                startActivityForResult(new Intent(Intent.ACTION_VIEW,
```
lakshyagupta21: What if this also results into an Exception, have you thought about that case?

naman653: This exception handling was already there, I just changed the method to start an activity. Also, why would this result in an exception? As the browser is present in every android device, so it will definitely behave as intended.

PR #344, opened 2020-01-22
feat: long press on filled forms navigate to ODK collect

### LINKED ISSUE #343: Navigate to collect when sending filled forms
Android Samsung J8

#### Problem description

Add a new feature on long pressing the filled form to direct to odk collect and show the form and I think no need add this feature to blank forms

#### How this feature helps to user

If user wants to review the form so, he can directly review from here instead of going to odk collect 

Not a problem a new feature

#### Expected behavior
By on long pressing the filled form should direct to odk collect with form ID

@opendatakit-bot claim

Chromicle:
Closes #343 

-  Checked whether collect launching as expected to a particular form by long pressing
-  Checked whether collect is launching as expected while reviewing the forms
-  Checked when we open the form in collect the selected forms are got unselected or not, they are working as expected
-  Checked the device by changing the orientation 
-  Checked transferring forms working perfectly or not with this changes 

I removed the existing method to launch collect and wrote new class `LaunchCollect` in that I gave the method `openFormInCollect` to launch collect as we are using the same method in both the classes and I used the dependency injection for the same 

If user wants to check the form before he starts transforming he has to go the collect and check the form by, adding this feature user can directly check/review the form by long pressing the particular form

### DIFF
--- a/skunkworks_crow/src/main/java/org/odk/share/injection/config/AppModule.java
+++ b/skunkworks_crow/src/main/java/org/odk/share/injection/config/AppModule.java
@@ -15,6 +15,7 @@ import org.odk.share.rx.schedulers.BaseSchedulerProvider;
 import org.odk.share.rx.schedulers.SchedulerProvider;
 import org.odk.share.services.ReceiverService;
 import org.odk.share.services.SenderService;
+import org.odk.share.views.customui.LaunchCollect;
 
 import dagger.Module;
 import dagger.Provides;
@@ -65,6 +66,11 @@ class AppModule {
     @Provides
     TransferDao provideTransferDao(Context context) {
         return new TransferDao(context);
+    }
+
+    @Provides
+    LaunchCollect provideLaunchCollect(Context context) {
+        return new LaunchCollect(context);
     }
 
     @Provides
--- a/skunkworks_crow/src/main/java/org/odk/share/views/customui/LaunchCollect.java
+++ b/skunkworks_crow/src/main/java/org/odk/share/views/customui/LaunchCollect.java
@@ -1,0 +1,35 @@
+package org.odk.share.views.customui;
+
+import android.content.Context;
+import android.content.Intent;
+import android.net.Uri;
+import android.widget.Toast;
+
+import org.odk.share.R;
+
+/**
+ * @author by Chromicle (ajayprabhakar369@gmail.com)
+ * @since 1/21/2020
+ */
+
+
+public class LaunchCollect {
+    private Context context;
+
+    public LaunchCollect(Context context) {
+        this.context = context;
+    }
+
+    public void openFormInCollect(long instanceId) {
+        Intent intent = new Intent(Intent.ACTION_VIEW);
+        intent.setData(Uri.parse("content://org.odk.collect.android.provider.odk.instances/instances/" + instanceId));
+        intent.putExtra("formMode", "viewSent");
+        if (intent.resolveActivity(context.getPackageManager()) == null) {
+            Toast.makeText(context, context.getString(R.string.collect_not_installed), Toast.LENGTH_LONG).show();
+        } else {
+            intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+            context.startActivity(intent);
+        }
+    }
+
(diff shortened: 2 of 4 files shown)


huangyz0918:
Thanks for this new feature! @Chromicle . But everytime we introduce new features, you'd better send a post to [ODK Forum](https://forum.opendatakit.org/) and seek for feedback. 

And for this feature, how can our user know long pressing can launch the Collect? If you are the product manager, how can you design and indicate your users about this?

Chromicle:
> Thanks for this new feature! @Chromicle . But everytime we introduce new features, you'd better send a post to [ODK Forum](https://forum.opendatakit.org/) and seek for feedback.

As this is the small feature I feel that no need to post in the forum, I will do from the next time 
 
> And for this feature, how can our user know long pressing can launch the Collect? If you are the product manager, how can you design and indicate your users about this?

There are so many sharing platforms like [ShareIt](https://play.google.com/store/apps/details?id=com.lenovo.anyshare.gps&hl=en_IN) and [MiDrop](https://play.google.com/store/apps/details?id=com.xiaomi.midrop&hl=en_IN) do the same so, I feel that user can know about this feature and we can include about this feature in `about app` so, user can know about this feature

lakshyagupta21:
> As this is the small feature I feel that no need to post in the forum, I will do from the next time

@Chromicle I would suggest to not directly start working on any new feature until it is been discussed within the community. And I agree that every feature need not be discussed on forum until and unless it is impacting large set of users.


> There are so many sharing platforms like [ShareIt](https://play.google.com/store/apps/details?id=com.lenovo.anyshare.gps&hl=en_IN) and [MiDrop](https://play.google.com/store/apps/details?id=com.xiaomi.midrop&hl=en_IN) do the same so, I feel that user can know about this feature and we can include about this feature in `about app` so, user can know about this feature

I agree there are so many apps in market doing lot of things, but we should not do the same thing they're doing until it is really required. For taking inspiration it is completely fine but having each and every feature is not required I guess. And I agree with @huangyz0918 that features should be discussed first then we should go for the implementation. And I don't think including everything in `About app` would be good way to help user, Ideally no one really reads `About app` but we still keep it so that user can know what this app actually does, we can't include "how to and what to" do in the app in `About app`.

Chromicle:
Yeah, @lakshyagupta21 @huangyz0918   from now I will discuss the new feature before I work on implementation I have made a thread on [slack](https://opendatakit.slack.com/archives/CAM2Z7UBD/p1579764096002000) please give your opinions

lakshyagupta21:
Can we add a long press action on any item when list with checkbox is there. @Chromicle What do you think? Or if you read any article or some Google Material Design guidelines or anywhere else who recommended to implement a long press on list with checkbox, can you share it?

Chromicle:
Yeah, we can add the long-press action while using checkbox as if we long press on the list 
- Functionality of checkbox will not get any change 
- Selected forms will not get unchecked if we open any form 
- As we already used onclick functionality for selecting forms and I think the best approach for this is long press

I took google Gmail as a reference and I did, but in Gmail, we have star the mail and we can select mails with long press \
mail

I got this idea to implement this feature if we have any better suggestions I will implement that

lakshyagupta21:
> I took google Gmail as a reference and I did, but in Gmail, we have star the mail and we can select mails with long press

Star is used for individual items and its not grouping them into one so that it can do some operation on the selected star items, if we talk about the checkboxes then we ask user to select them and then we apply some operation on them as a group which results into something useful depending on the user-chosen option. If you're seeing it now, that when you long-press the item it starts creating a list for the checked and unchecked items so that users can at last select either to delete them or mark them as unread/read.

Chromicle:
> Star is used for individual items and its not grouping them into one so that it can do some operation on the selected star items

I understood I feel that long press will be the good one to open form 
@lakshyagupta21 @huangyz0918 if you have any other better approach for this I will implement

for checkbox, this is one more reference, this app is [shareIt](https://play.google.com/store/apps/details?id=com.lenovo.anyshare.gps&hl=en_IN)(one file sharing platform) 
they are following the same pattern with checkbox to choose file 
shareit

Issue #136, opened 2019-02-28
Change the default hotspot name


lakshyagupta21:
#### Software and hardware versions 
Devices having Android version < 8
#### Problem description
For devices having android version < 8, by default hotspot name is set to ODK-SHARE-odk-share which should be changed to ODK-SKUNKWORKS-odk-skunkworks.
#### Steps to reproduce the problem
1. Press 'Send'
2. Select the forms
3. Press 'Send'
4. Start Hostpot
5. Default name would show ODK-SHARE-odk-share
#### Expected behavior
Default name should be changed to ODK-SKUNKWORKS-odk-skunkworks

prajurock:
@lakshyagupta21 i would like to work

prajurock:
@opendatakit-bot claim

shobhitagarwal1612:
@lakshyagupta21 Is there a specific reason as to why we can't name the hotspot with a shorter name, like `skunkworks-crow`?

lakshyagupta21:
@shobhitagarwal1612 We can have skunkworks-crow. My idea is to have something in prefix every time a user creates a hotspot so that nearby devices can uniquely identify the network created by the skunkwork-crow app.

PR #348, opened 2020-01-30
Def name

### LINKED ISSUE #347: Disable button instead of showing toast for null validation
Android 9, Samsung J8

#### Problem description
Currently, In settings activity we are showing the toast that `Name should not be empty`  if the length of the string equals to 0 in every `EditTextPreference`
Instead of showing toast, disable the positive button so, user can directly know that he did not have to enter the null value in EditText or else he/she have to click once again on preference and have to edit 

Go to Settings activity
Open any editText and give a null value 
observe the toast
 
#### Expected behavior
Instead of showing toast disable the positive button for editTextPreference for null validation 

@opendatakit-bot claim

Chromicle:
Closes #347 #136  

- Checked whether the positive button is disabling when the length of the text is null
- Checked any new warnings are producing or failing any tests, working perfectly as expected
- Checked by changing the orientation 
- Checked the default name for the hotspot is device name or not 

I added the new custom preference called `CustomEditTextPreference` and I implemented the editTextWatcher to check for every text change \
In the method of `onEditTextChanged` checked if the length of the editText is zero then I am disabling the positive button of the alert dialog, we can implement the any further features if we want in  `CustomEditTextPreference`  and I wrote the tests according to that \
I made the one more change that I changed the `CheckBoxPreference` to `SwitchPreference`  as according to [material design](https://material.io/components/selection-controls/#switches) switches are used to toggle the state of a single setting on or off and checkboxes are used for select one or more items from a set like selecting the `forms`

For default name of hotspot as @shobhitagarwal1612 suggested I made that as `skunkworks-crow deviceName` in my case it is `skunkworks-crow SM-J810G` 

By adding this feature user can directly know that he did not have to give the `null` value in edit text or else he has to select the preference once again then he has to edit and save 
By making default hotspot name as the device name so it will be like unique 

#### GIF
ezgif com-video-to-gif (2)


skunkworks_crow/src/main/java/org/odk/share/views/ui/settings/CustomEditTextPreference.java
```diff
@@ -0,0 +1,73 @@
+package org.odk.share.views.ui.settings;
```
huangyz0918: We should use some names that can indicate the class usage, not just `CustomEditTextPreference` @Chromicle .

skunkworks_crow/src/main/java/org/odk/share/views/ui/settings/CustomEditTextPreference.java
```diff
@@ -0,0 +1,73 @@
...
+import android.text.Editable;
+import android.text.TextWatcher;
+import android.util.AttributeSet;
+import android.widget.Button;
+
+/**
+ * @author by Chromicle (ajayprabhakar369@gmail.com)
+ * @since 1/30/2020
+ */
+public class CustomEditTextPreference extends EditTextPreference implements DialogInterface.OnClickListener {
```
huangyz0918: I'm not sure do we really need a custom preference for this tiny check? @Chromicle 

Can we use some method like,

```java
private void checkNullEditTextPreference(EditTextPreference edttxtpref) {
    edttxtpref.setOnPreferenceChangeListener(new OnPreferenceChangeListener() {
        @Override
        public boolean onPreferenceChange(
                android.preference.Preference preference, Object newValue) {
            if (newValue.toString().trim().equals("")) {

              // Get Dialog and disable button here

                return false;
            }
            return true;
        }
    });
}
```
and receiving a EditTextPreference? You know, building a custom class is always a heavy work and not good for code reading and code refactoring. I didn't test for this method but you can have a try.

Chromicle: As far as I know, `setOnPreferenceChangeListener` will be triggered when we press the positive button of the dialog so there is no use if we disable the button then so, here we want to check the edit text for every time so we have to implement the `editText watcher`  correct me if I am wrong 

Now I updated the code without implementing the `customEditTextPreference` can please have a look now and thanks for your suggestions 😄

skunkworks_crow/src/main/java/org/odk/share/views/ui/settings/CustomEditTextPreference.java
Chromicle: I gave that name because maybe it will be useful for further implementations of any features

skunkworks_crow/src/main/res/values/strings.xml
```diff
@@ -184,4 +181,6 @@
     <string name="finalized_on_date_at_time">\'Finalized on\' EEE, MMM dd, yyyy \'at\' HH:mm</string> <!-- http://docs.oracle.com/javase/6/docs/api/java/text/SimpleDateFormat.html -->
     <string name="sent_on_date_at_time">\'Sent on\' EEE, MMM dd, yyyy \'at\' HH:mm</string> <!-- http://docs.oracle.com/javase/6/docs/api/java/text/SimpleDateFormat.html -->
     <string name="sending_failed_on_date_at_time">\'Sending failed on\' EEE, MMM dd, yyyy \'at\' HH:mm</string> <!-- http://docs.oracle.com/javase/6/docs/api/java/text/SimpleDateFormat.html -->
+
+    <string name="one_space">\u0020</string>
```
lakshyagupta21: There is no need to create a string resources for space.

skunkworks_crow/src/main/java/org/odk/share/views/ui/settings/SettingsActivity.java
```diff
@@ -72,16 +74,20 @@ private void addPreferences() {
...
         odkDestinationDirPreference = (EditTextPreference) findPreference(PreferenceKeys.KEY_ODK_DESTINATION_DIR);
 
         prefs = PreferenceManager.getDefaultSharedPreferences(getApplicationContext());
 
         defaultMethodPreference.setSummary(prefs.getString(PreferenceKeys.KEY_DEFAULT_TRANSFER_METHOD,
                 getString(R.string.default_hotspot_ssid)));
-        hotspotNamePreference.setSummary(prefs.getString(PreferenceKeys.KEY_HOTSPOT_NAME,
-                getString(R.string.default_hotspot_ssid)));
+
+        String defaultHotspotName = getString(R.string.app_name) + getString(R.string.one_space) + Build.MODEL;
```
lakshyagupta21: If this should just be a device odk-skunkworks-crow or the device name. Not a combination of both because if try to keep both the combinations, then it may get bigger for few devices then it may end up truncating or creating a double line for displaying just the name of the bluetooth/hotspot device.

Chromicle: so, can we go with `device odk-skunkworks-crow`

Chromicle: or can we go with `device name`

PR #350, opened 2020-02-02
Added App information in about section

### LINKED ISSUE #175: Add app info in About
#### Problem description
There is no information about the current app info in About screen

Goto options menu > About

#### Expected behavior
Above `Open Source Licenses`, a new entry should be present with the current app info

devanshi7799:
Closes #175 

I ran the application on my phone and it worked perfectly fine

Here, I have added a new html file for app information in the assets and made some changes in AboutActivity.java

New html page is added regarding the information about the application as it was expected

### DIFF
--- a/skunkworks_crow/build.gradle
+++ b/skunkworks_crow/build.gradle
@@ -74,4 +74,5 @@ dependencies {
     // Permissions Dispatcher
     implementation "org.permissionsdispatcher:permissionsdispatcher:4.5.0"
     annotationProcessor "org.permissionsdispatcher:permissionsdispatcher-processor:4.5.0"
+    implementation 'androidx.browser:browser:1.2.0'
 }
--- a/skunkworks_crow/src/main/java/org/odk/share/views/ui/about/AboutActivity.java
+++ b/skunkworks_crow/src/main/java/org/odk/share/views/ui/about/AboutActivity.java
@@ -1,11 +1,14 @@
 package org.odk.share.views.ui.about;
 
 import android.content.Intent;
+import android.net.Uri;
 import android.os.Bundle;
 import android.view.View;
 
 import androidx.appcompat.app.AppCompatActivity;
 import androidx.appcompat.widget.Toolbar;
+import androidx.browser.customtabs.CustomTabsIntent;
+import androidx.core.content.ContextCompat;
 import androidx.recyclerview.widget.LinearLayoutManager;
 import androidx.recyclerview.widget.RecyclerView;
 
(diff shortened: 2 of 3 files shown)


devanshi7799:
ezgif com-video-to-gif

skunkworks_crow/build.gradle
```diff
@@ -74,4 +74,5 @@ dependencies {
     // Permissions Dispatcher
     implementation "org.permissionsdispatcher:permissionsdispatcher:4.5.0"
     annotationProcessor "org.permissionsdispatcher:permissionsdispatcher-processor:4.5.0"
+    implementation 'androidx.browser:browser:1.2.0'
```
huangyz0918: Why introducing a third-party lib for this small feature @devanshi7799 ? If I were you, I will implement that using a simple web view. 

Downsides for introducing a new lib in this case are:
1. Some third-party libs have many redundant functions, that will increase the apk size. 
2. Some third-party libs are not well tested that may introduce bugs that cannot fix.

So we are trying not to use third-party libs unless we have to do that, I think this would be helpful for you.

devanshi7799:
@huangyz0918 should i do it using a webview?

huangyz0918:
> @huangyz0918 should i do it using a webview?

Yes, you should have a try.

skunkworks_crow/build.gradle
lakshyagupta21: > Why introducing a third-party lib for this small feature @devanshi7799 ? If I were you, I will implement that using a simple web view.

Hi @huangyz0918, it's a custom tabs dependency, and nowadays every app has in-app browser feature which provides a look and feel of the browser within the app without actually moving out of the app, so I think its fine we even have this in ODK-Collect but not AndroidX dependency. I think @devanshi7799 this is fine no need to explore the webview as of now, if any feature ask comes in future then we can pick that up or if it we face any issues due to this then we can pick this up.

skunkworks_crow/src/main/java/org/odk/share/views/ui/about/AboutActivity.java
```diff
@@ -47,19 +51,25 @@ protected void onCreate(Bundle savedInstanceState) {
...
         adapter.addItem(new AboutItem(R.string.user_guide, R.drawable.ic_stars));
         recyclerView.setAdapter(adapter);
     }
 
-    @Override
+            @Override
     public void onItemClick(View view, int position) {
         if (position == 0) {
+
+            new CustomTabsIntent.Builder()
```
lakshyagupta21: My main concern was that what if the device doesn't support custom tabs, it would be easier for you to understand if you can explore ODK-Collect Custom tabs implementation that they have because that handles lot of things which we might not be able to test on our devices, there are even few devices which doesn't have browser in them, so this change may crash the app on those devices.

devanshi7799: @lakshyagupta21 Should i make changes according to odk-collect custom tabs?

PR #356, opened 2020-02-04
fix: MainActivity UI is updating after updating in database

### LINKED ISSUE #309: After form transfer status on Main screen is not updated correctly

#### Problem description
Whenever any instance form transfer happens, there is some discrepancy in reviewed and unreviewed status on the main screen.
1. Download all the available forms(at least two is needed to distinguish the status) from the default server on both sender and receiver.
2. Create one instance of Collect-Location form in one device and send it to the other.
3. Go on home screen and check the status of unreviewed form, it shows the unreviewed counter of `All widgets` form.
#### Expected behavior
It should show the unreviewed status for Collect-location form.
#### Other information 
There is some code problem in FormsAdapter class where we're querying the instances and transfer table.

Chromicle:
Closes #309 

- Checked whether `reviewed` and `unreviewed` count is getting perfectly or not 
- Checked whether the count is updating when on pressing the back button 
- Checked whether any new warnings are produced by adding this code
- Checked any tests are failing
- Checked by changing the orientation of the device

Removed previous approach of the query(checking every time status of form) and added the query for getting `receivedCount` and `reviewedCount` so, we can get the unReviewed count with `receivedCount-reviewedCount` and Updated the forms adapter when user presses back button from another activity in `onResume`

By updating the UI in `onResume` so, the user no need to restart the activity and get the change 

#### GIF
ezgif com-video-to-gif (3)

### DIFF
--- a/skunkworks_crow/src/main/java/org/odk/share/views/ui/main/MainActivity.java
+++ b/skunkworks_crow/src/main/java/org/odk/share/views/ui/main/MainActivity.java
@@ -110,7 +110,6 @@ public class MainActivity extends FormListActivity implements LoaderManager.Load
     @Override
     public void onResume() {
         super.onResume();
-        //update the list
         if (formAdapter != null) {
             formAdapter.notifyDataSetChanged();
         }
--- a/skunkworks_crow/src/main/java/org/odk/share/views/ui/main/MainActivity.java
+++ b/skunkworks_crow/src/main/java/org/odk/share/views/ui/main/MainActivity.java
@@ -107,6 +107,15 @@ public class MainActivity extends FormListActivity implements LoaderManager.Load
         addListItemDivider();
     }
 
+    @Override
+    public void onResume() {
+        super.onResume();
+        //update the list
+        if (formAdapter != null) {
+            formAdapter.notifyDataSetChanged();
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


huangyz0918:
Thanks @Chromicle , this is an important fix. The code looks good to me but I need sometime to test and make sure this fix is valid and safe.

skunkworks_crow/src/main/java/org/odk/share/views/ui/main/MainActivity.java
```diff
@@ -107,6 +107,15 @@ protected void onCreate(Bundle savedInstanceState) {
         addListItemDivider();
     }
 
+    @Override
+    public void onResume() {
+        super.onResume();
+        //update the list
```
huangyz0918: The code is very clear and we don't need this comment.

Chromicle: Yeah, I updated.
can you please have a look at my previous [PR's](https://github.com/opendatakit/skunkworks-crow/pulls/Chromicle)

huangyz0918:
@Chromicle Thanks a lot! I'm always looking for the day that this issue can be addressed, it's you!

PR #374, opened 2020-03-21
Fix: Add Toast "Press BACK to exit" for exit

### LINKED ISSUE #372: Feat: Add exit confirmation dialog
All

#### Problem description
While at home screen, if back button is pressed, the app directly closes.
Open the app and click back.
#### Expected behavior
Upon adding the feature, a dialog box will be displayed which will ask user to confirm wether to close the app or back was pessed accidentlly.

pritsahkar:
Closes #372 

I tested it on Android 9.0.

As comments in this issue, I figure out that exit confirmation is good, but not by dialog so I'm doing this by an easy approach.

No regression risks

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,7 +7,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.5.1'
+        classpath 'com.android.tools.build:gradle:3.6.1'
         
 
         // NOTE: Do not place your application dependencies here; they belong
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
-#Mon Oct 28 17:15:18 IST 2019
+#Mon Mar 16 16:33:49 IST 2020
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.4.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.4-all.zip
--- a/skunkworks_crow/src/main/java/org/odk/share/views/ui/main/MainActivity.java
+++ b/skunkworks_crow/src/main/java/org/odk/share/views/ui/main/MainActivity.java
@@ -9,6 +9,7 @@ import android.database.Cursor;
 import android.net.Uri;
 import android.os.Build;
 import android.os.Bundle;
+import android.os.Handler;
 import android.view.Menu;
 import android.view.MenuItem;
 import android.view.View;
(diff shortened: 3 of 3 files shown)


pritsahkar:
@Chromicle can you help me I don't get this Circleci work?

Chromicle:
There are some checkstyle violation to solve that please run `./gradlew checkcode` locally with that you will get a detailed explanation to solve that error 

I don't know this is a valid approach or not, I am just giving you a suggestion to solve that error 
@huangyz0918 can you give your opinion on this approach(double press to exit)

build.gradle
```diff
@@ -7,7 +7,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.5.1'
```
huangyz0918: The file should not be change, since this is another issue to update the gradle.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Mon Oct 28 17:15:18 IST 2019
+#Mon Mar 16 16:33:49 IST 2020
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.4.1-all.zip
```
huangyz0918: The same above.

skunkworks_crow/src/main/java/org/odk/share/views/ui/main/MainActivity.java
```diff
@@ -292,4 +294,22 @@ private void setUpLoader() {
             }
         }
     }
+
+    // Add toast "Press BACK to exit"
```
huangyz0918: The comment is self-explanatory, no need here.

skunkworks_crow/src/main/java/org/odk/share/views/ui/main/MainActivity.java
```diff
@@ -292,4 +294,22 @@ private void setUpLoader() {
...
+    @Override
+    public void onBackPressed() {
+        if(BACK_FLAG){
+            super.onBackPressed();
+            return;
+        }
+        BACK_FLAG = true;
+        Toast.makeText(getApplicationContext(),"Press BACK to exit",Toast.LENGTH_SHORT).show();
```
huangyz0918: `Press BACK to exit` should be something like `Press BACK again to exit program`, and put the string in `string.xml`.

skunkworks_crow/src/main/java/org/odk/share/views/ui/main/MainActivity.java
```diff
@@ -292,4 +294,22 @@ private void setUpLoader() {
...
+
+        new Handler().postDelayed(new Runnable() {
+            @Override
+            public void run() {
+                BACK_FLAG = false;
+            }
+        },2000);
```
huangyz0918: If the delay time `2000` is static, is better to use a static variable to present.

pritsahkar:
@huangyz0918 Thanks for the suggestion, I make changes and make a new pull request

@Chromicle I run ./gradlew checkout but It still gives me error- "Do not use Windows line endings" in all files even in all lines

huangyz0918:
> @huangyz0918 Thanks for the suggestion, I make changes and make a new pull request

I'm a bit confused why you need a new PR? @pritsahkar

pritsahkar:
> > @huangyz0918 Thanks for the suggestion, I make changes and make a new pull request
> 
> I'm a bit confused why you need a new PR? @pritsahkar

In this PR, my commit is in an unknown repository, now I try to add commit but this time code is in the master branch so this forces me to new PR and apologizes me for a push into the master branch

